PR #137, opened 2025-11-28
feat: replace language toggle with Select component and support React

shangxinxiaolin:


guangliang2019:
Sorry, that won’t work. You can’t add React as a dependency — you must implement this using Astro and plain HTML/CSS/JS.

PR #183, opened 2026-04-04
feat+fix+docs: transition v0 self-transition contract and preview pipeline

cyjin-yl:
this pr bundles the transition contract implementation, www preview pipeline enhancements, demo hardening, and follow-up fixes.

what's inside:

- `feat(transition,adapter)`: implement `astransition` and cross-runtime adapter support. base-transition handles the single-unit presence lifecycle: `closed → entering → entered → leaving → closed`.
- `feat(demo)`: introduce cross-runtime `demoruntimeapi` and transition demos.
- `fix`: resolve post-rebase hook rename regression and synthetic click re-entrancy.
- `fix(demo)`: use double `raf` in react demo renderer to guarantee dom refs on initial mount.
- `docs(www)`: split hover-card docs into standalone base/shadcn pages.
- `docs(www)`: add code preview panels to base/shadcn prototype pages.
- `feat(www)`: auto-resolve `codemap` by `demoid` and unify runtime switching in `previewer`.
- `docs(www)`: remove duplicate code blocks and add css for transition code previews.
- `docs(www)`: add button interactions to base-transition codemap previews (wc/react/vue).
- `fix`: address copilot review feedback on pr #183 — scope demo renderer caches per adapter instance, fix copy button rebinding after runtime switch, pin `packagemanager`.
- `fix(www)`: restore visible scrollbar in code panel expanded state.
- `refactor`: per review feedback and prototype boundary whitepaper, narrow `astransition` to the self-transition abstraction only. revert `skipself` context framework changes and remove nested transition coordination from this pr.

commits:
`f8f3987`, `e469a75`, `5fc9919`, `2ab2ba6`, `1b03ed9`, `732b4ef`, `37827f5`, `0787b1c`, `b5a9fc0`, `b96e6a2`, `d18e969`, `9610d98`


apps/www/src/content/docs/en/prototypes/base/transition.mdx
```diff
@@ -0,0 +1,124 @@
+---
```
guangliang2019, line 1: Well-structured and comprehensive component documentation, including an API table—much appreciated. This will serve as a valuable reference for writing other pages.

guangliang2019:
Thanks for pushing this forward. There are several parts of this PR that I think are moving in a good direction:

- modeling transition as presence-state governance rather than tying it directly to a single platform animation mechanism
- keeping completion as an adapter/host-driven signal via `complete()`
- thinking about interruption strategies (`reverse` / `wait` / `immediate`) early, instead of only supporting the ideal path
- wiring the feature into docs/demo quickly so the capability becomes visible instead of staying purely abstract

My main concern is not with those directions themselves, but with the current boundary of `asTransition`.

At the moment, this PR seems to fold two different concerns into the same abstraction:

1. a **self transition** hook that governs the presence lifecycle of a single unit
2. a **nested coordination** mechanism where parent/child transitions coordinate leave timing

I think these should stay separate.

My default expectation is that `asTransition` should remain focused on its own lifecycle only. In the common case, its content simply enters and leaves together with it. If some inner content has its own timing requirements or exit ordering, that inner part is no longer just passive content — it becomes its own interaction participant. At that point, the problem starts to look more like a `transition-group` / `transition-child` split (or a root/node split), rather than a single base transition hook with extra coordination logic.

Because of that, I would suggest the following adjustments for this PR:

- keep `asTransition` scoped to the single-unit / self-transition case
- do not merge nested transition coordination into the base transition abstraction yet
- roll back the `context` API option change for now

On the `context` side: I do think the idea behind `skipSelf` is valuable. It reveals a real gap for cases where a prototype may need to be both a provider and a consumer of the same context while intentionally bypassing its own provider. That seems like an important case for things such as future tree-like prototypes. But since Proto UI is intentionally conservative about core API changes, I do not think this PR is the right place to expand the `context` surface yet. It would be better to revisit that later in a more explicit and formal design discussion.

So overall, my suggestion would be:

- keep the transition state machine direction
- keep the adapter/host-driven completion model
- keep the interrupt strategy work
- but narrow this PR back to the self-transition abstraction, and leave group/child-style coordination for a separate design

I think that would make the contract boundary much clearer while still preserving the good work already done here.

guangliang2019 (commented):
我觉得这次 PR 里有几条方向本身是好的，尤其是：

- 把 transition 抽象成 presence-state governance，而不是直接绑死在某个平台动画实现上
- 通过 `complete()` 把“完成信号”留给 adapter / host 驱动，这个职责切分是合理的
- 提前考虑 interrupt 策略，说明设计是在面对真实交互抖动场景
- docs/demo 及时跟进，也有助于让能力形成可感知的闭环

主要需要调整的不是这些方向本身，而是这次把“单体 transition”与“嵌套 transition 协调”折进了同一个 abstraction 里。

我更倾向于这次先把 `asTransition` 收缩成单体 presence hook，只处理自身状态机；
如果以后要支持 group / child / boundary 一类的嵌套协调，再作为更明确的角色来设计。

另外，`context` 这次先不要引入 `options` 形态的 core API 变更。这个思路是有价值的，也暴露了 Proto UI 在“同时是 provider 和 consumer，且需要有选择地越过自身 provider”的场景里确实还有缺口，但这部分更适合以后单独、正式地设计。

255doesnotexist:
@guangliang2019 done. force-pushed a clean branch.

what changed:
- narrowed `asTransition` to a single-unit presence hook. stripped out `dependsOnParentTransition`, `TRANSITION_CONTEXT`, and child registries. state machine now only manages `closed → entering → entered → leaving → closed` for one element.
- reverted `skipSelf` core changes entirely. `packages/core`, `packages/modules/context`, and `packages/runtime` are back to `main` state via `git checkout main --`. no framework-wide api expansion in this pr.
- added a **v0 scope note** at the bottom of both zh-CN and EN transition docs. nested exit coordination (group / child / boundary) is explicitly called out as a future separate abstraction, not part of v0.

let me know if anything else needs a tweak.

guangliang2019 (commented):
I think there is one more design decision worth making explicitly before this goes further:

If we want transition to expose a configuration that lets the maker specify whether the element should be merely visually hidden or actually removed before enter / after leave, then we should decide whether `asTransition` is still a regular surface-level asHook, or whether it should be upgraded into a privileged hook path.

I think there are two coherent directions here.

## Direction A: keep `asTransition` as a regular hook

In this direction, `asTransition` stays as a self-contained presence-state machine at the prototype surface:

- it governs `closed -> entering -> entered -> leaving -> closed`
- it may expose configuration about desired presence policy
- but it does **not** rely on adapter / host-specific cooperation for disposal or delayed actual removal

If we choose this direction, then we should be strict about the boundary:

- no host-side disposal dependency should be designed into the prototype surface
- the prototype should not leave implicit “adapter hooks” for delayed unmount behavior
- adapter code should not be expected to special-case transition disposal semantics, because that would already cross the architecture boundary

In that model, “hidden vs absent” must stay within what the surface abstraction itself can honestly represent. If actual delayed removal is out of reach without system capabilities, then the transition API should avoid pretending it controls that part.

This is the simpler route, but it also means `leaveDuration` / `enterDuration` should remain weaker values (at most descriptive timing inputs, not stronger lifecycle guarantees), because there is no formal capability path for consuming them as part of real delayed disposal.

## Direction B: promote `asTransition` into a privileged hook path

If we want transition to support things like delayed actual removal after leave completes, then I think this should be treated as a privileged hook design.

At that point, transition is no longer only governing local state. It starts depending on runtime / adapter-side capabilities, because “actual removal later” is not something a normal surface hook can guarantee on its own.

If we choose this direction, I would suggest following the same architectural route as the hooks under `packages/hooks`:

- treat `asTransition` as a special hook that uses system-backed capabilities
- introduce the required capability path through modules / runtime / hook integration
- let adapters provide the concrete host-side ability needed for delayed actual removal
- keep the hook logic coupled to those capabilities through formal integration points, rather than by leaving implicit behavior for adapters to guess

This route is more complex to implement, but architecturally it is cleaner and more stable. It also gives `enterDuration` / `leaveDuration` a much more solid consumption path: instead of being vague interface values or timing hints, they can participate in a formal lifecycle that is actually backed by runtime / host cooperation.

## My suggestion

Both directions are valid, but I do not think the middle state is stable:

- if `asTransition` stays a regular hook, then host-disposal concerns should be explicitly kept out
- if `asTransition` needs delayed actual removal, then it should be upgraded into a privileged hook path instead of implicitly depending on adapter behavior

I think making that choice explicitly now will keep the transition design much cleaner.

255doesnotexist:
@guangliang2019 choosing **direction b** — upgrading `asTransition` into a privileged hook path.

design summary:
- introduce a new privileged module `@proto.ui/module-presence` that handles structural mount/unmount timing.
- `transition` becomes a specialization of `presence`: `asTransition` uses `getActiveAsHookContext` + privileged registration to access the presence module's facade/port.
- `packages/runtime/src/instance/execute/with-host.ts` gets async interception points for both mount and unmount, driven by `module-presence` policies.
- adapters (react/vue/wc) wire a `hostBridge` cap into the module, providing the concrete host-side ability for delayed actual removal and pre-enter DOM creation.
- state machine stays in `asTransition` but delegates structural decisions to `presence` via intent signals (`enter` / `leave`), keeping the boundary clean.

this lines up with the `packages/hooks` privilege pattern and gives `enterDuration` / `leaveDuration` a formal lifecycle backed by runtime/host cooperation.

starting implementation now. will push in follow-up commits.

255doesnotexist:
hi @guangliang2019,

went with direction b: upgraded `asTransition` to a privileged hook backed by a new system-level `@proto.ui/module-presence` module. this gives it formal control over structural mount/unmount timing across all adapters (wc, react, vue).

what changed:
- `as-transition` is now a privileged hook using `getActiveAsHookContext` + `rt.register(..., { privileged: true })`
- created `@proto.ui/module-presence` with `PresenceModuleImpl` managing the phase machine `absent → mounting → present → unmounting`
- intercepted mount/unmount in `runtime/src/instance/execute/with-host.ts` via `awaitMount()` and `awaitUnmount()`
- wired presence bridges into all three adapters:
  - react: `setShouldExist` toggles root render
  - vue: `v-if` tied to presence state
  - wc: defers `disconnectedCallback` until unmount is approved
- fixed a mount deadlock when starting from `closed` without `open=true` — `leave` intent from `absent` now resolves pending mounts correctly
- updated en/zh docs to explain the architecture

tests from this refactor:
- `as-transition` contract: 18/18 passing
- `presence` module unit: 5/5 passing
- runtime integration: 2/2 passing
- adapter suites: wc 102/102, react 29/29, vue 30/30 passing

there are 14 remaining failures on this branch, but they are **not** pre-existing from `main`. i bisected and found they were introduced earlier on `feat/transition` itself:

- **culprit commit**: `9610d98` ("refactor: narrow asTransition to self-transition only and revert skipSelf")
- **culprit file**: `packages/runtime/src/kernel/handles/run.ts`
- **root cause**: the entire `run.anatomy.order` sub-object was accidentally removed while reverting `skipSelf`. this breaks `as-collection-item.ts` and any code that calls `run.anatomy.order.version(...)`.
- **verification**: `879c46c` (parent of `9610d98`) passes all 30 affected tests; `9610d98` fails them.

commits are all cleaned up to match the repo's conventional commit style. pushed to `feat/transition-work`. let me know if anything needs adjusting before merge.

guangliang2019 (commented):
I have not confirmed the expected changes on the current PR branch yet.

Also, there is now an extra branch in the repo: feat/transition-work (instead of the current feat/transition).

You can:

submit a new PR from the branch with the highest completion and close this PR, or
merge the feat/transition-work contents back into the current branch, so this PR reflects the actual implementation being reviewed.

packages/prototypes/base/src/transition/as-transition.ts
```diff
@@ -0,0 +1,314 @@
+// packages/prototypes/base/src/transition/as-transition.ts
+import type { RunHandle } from '@proto.ui/core';
+import { getActiveAsHookContext } from '@proto.ui/core/internal';
```
guangliang2019: asTransition 提升为特权 hook 之后，应该放在 hooks 包中而非原型库中，后面我会考虑做更干净的架构隔离，让 getActiveAsHookContext 没法从 core 导入

guangliang2019: 但目前不要替我进行架构调整，只需要将该 hook 移动到 hooks 包内就行

packages/prototypes/base/src/transition/as-transition.ts
```diff
@@ -0,0 +1,314 @@
...
+    defaultOpen: { type: 'boolean', empty: 'fallback' },
+    appear: { type: 'boolean', empty: 'fallback' },
+    enterDuration: { type: 'number', empty: 'fallback' },
+    leaveDuration: { type: 'number', empty: 'fallback' },
+    interrupt: {
+      type: 'string',
+      empty: 'fallback',
+      enum: ['reverse', 'wait', 'immediate'],
+    } as any,
+    onBeforeEnter: { type: 'any', empty: 'accept' },
```
guangliang2019: 实际上，Proto UI 的 props 是不支持函数的，正确的做法是用 def.expose.event 定义一个事件，然后在你希望的该事件发生的时刻 run.event.emit 它，这样

guangliang2019: 这样，它在 React 中会被翻译为 onEvent 的 props，而在 Vue 中被翻译为 emit 自定义事件 + onEvent 的 props。

guangliang2019 (commented):
临近收尾，加油

255doesnotexist:
@guangliang2019

following up with the latest progress since my last update. this branch has moved forward 11 commits and is ready for another review pass.

the privileged hook migration is now complete. asTransition is fully implemented as a privileged hook backed by the new @proto.ui/module-presence module. structural mount and unmount timing is coordinated across all three adapters (React, Vue, and Web Component) via formal host bridges rather than adapter-level guesses.

on the adapter side, the soft-unmount and remount pipeline is also fixed end to end. React and Vue adapters now keep the element in the DOM during the closed state using CSS hiding instead of physically removing it. this gives the correct remount baseline for CSS transitions. i also added a double-RAF guarantee to ensure the data-transition-state="closed" baseline is painted for at least one frame before soft-unmount proceeds. the actual soft-unmount is deferred by one frame so that a rapid re-entry does not cancel the animation before it starts. during soft-unmount, exposes remain visible so imperative controls like controls.enter and controls.leave stay callable.

for tests, i added RAF-frame unmount regression tests for the React and Vue adapters and improved the React closed-enter test to cover reverse re-entry cancel. the 14 failures i reported earlier have been fixed. they were caused by the accidental removal of run.anatomy.order in commit 9610d98, and that has been restored.

docs were updated too. the transition lifecycle rules now use expose events (beforeEnter, afterEnter, beforeLeave, afterLeave). the EN and ZH docs also clarify the advisory physical removal behavior for React and Vue.

in the latest commit, i ran a simplify review pass and extracted shared utilities to reduce duplication. createSoftUnmountScheduler was moved into @proto.ui/adapter-base and is now shared by the React and Vue adapters, removing about 150 lines of duplicated animation tracking logic. createInstanceTreeMarkers factory was moved into @proto.ui/adapter-base, eliminating the triplicated instance-tree.ts files across React, Vue, and Web Component. an isPromiseLike helper was added in packages/modules/presence/src/impl.ts, replacing five repeated Promise duck-typing checks. VALID_TRANSITIONS was hoisted to module scope in packages/hooks/src/as-transition.ts to avoid recreating the state guard table on every call.

current test status: React 29/29, Vue 30/30, Web Component 100+, all passing. the as-transition contract is 19/20. the single remaining failure (AS-TRANSITION-2100: rapid enter cancels pending unmount) is pre-existing on this branch and unrelated to recent changes.

the feat/transition-work side branch has been merged back into feat/transition, so everything in this PR reflects the actual implementation.

please take another look when you have time. happy to address any feedback.

guangliang2019 (commented):
代码非常棒，可以合入了，但在合入之前需要更新一下 lock file，vercel 因为 lockfile 策略比较保守没部署成功（不修的话，合入后官网不会自动部署）

255doesnotexist:
@guangliang2019 small follow-up: the pre-existing AS-TRANSITION-2100 failure mentioned in the last update is now fixed and pushed.

root cause: PresenceModuleImpl.setIntent('enter') in the unmounting phase unconditionally called bridge.mount(), even when bridge.unmount() had never been dispatched (two-stage leave, first stage). this caused an extra mount call on rapid re-entry.

fix: track pendingMount and pendingUnmount in PresenceModuleImpl. reverse interrupts now only call bridge.mount() or bridge.unmount() when the corresponding structural operation has actually been started. this makes the presence bridge contract consistent and eliminates the spurious second mount.

commit: e0f0107

current test status: all adapter, presence, and transition tests pass. no remaining failures on this branch.

255doesnotexist:
@guangliang2019 thanks for the review.

current status: all tests pass (141 files, 481 cases, 0 failures) on this branch. the latest commit 00d8dcc covers the react strictmode baseline fix and the presence stale-completion guard.

regarding the lockfile update you mentioned: i can run pnpm i locally and push the refreshed pnpm-lock.yaml if you need it in this pr. just let me know and i will add it.

255doesnotexist:
lockfile updated. pnpm-lock.yaml is now in sync with the workspace and pushed to this branch. everything is ready from my side.

255doesnotexist:
@guangliang2019 current status update:

1. lockfile updated: pnpm-lock.yaml refreshed and pushed (ce91e5e).
2. the vue "click twice to enter from closed" issue is fixed and pushed (a9b43d2). root cause was the same baseline race that react had: vue follow-up commits during the double-raf baseline were canceling the closed frame, so the first enter click got dropped. fixed by adding commitVersion + baseline-in-progress guard to the vue adapter, matching the react path.
3. demo-renderer callInScope now has a fallback for invokeInCallbackScope to avoid dropping first-click commands.
4. added a vue integration regression test in previewer-transition-command.demo.test.ts to lock the single-click enter behavior.

test result: 142 files passed, 482 tests passed, 0 failures.

from my side this branch is ready to merge. let me know if anything else is needed.

255doesnotexist:
@guangliang2019 

status: ready to merge from my side.

- lockfile updated and pushed (ce91e5e).
- vue "click twice to enter from closed" is fixed (a9b43d2). it was the same baseline race as react: follow-up commits during the double-raf baseline were dropping the first enter click. added commitVersion + baseline-in-progress guard to vue adapter.
- demo-renderer callInScope has a fallback so first-click commands are not lost when invokeInCallbackScope is exposed but not yet wired.
- added vue previewer integration regression test to lock single-click enter after closed.

tests: 142 files passed, 482 tests passed, 0 failures.

guangliang2019 (approved):
非常棒！asTransition 非常有用，辛苦了 ❤️

PR #184, opened 2026-04-06
feat(dialog): add dialog prototype with cross-runtime portal and z-index layering

cyjin-yl:
this pr closes issue #72.

changes:
- add base dialog family (root, trigger, overlay, content, title, description, close)
- add shadcn dialog skin on top of base parts
- fix body portal for react and vue by injecting `host` getter into `kernel.run`
- expose `kernel` on runtime session and add `host` to `RunHandle` type
- establish z-index layering: dialog uses `z-50`, hover-card and dropdown lowered to `z-40`
- regenerate tailwind prototype tokens
- add docs and demo pages for both base and shadcn variants
- add contract tests and adapter integration tests

ready for review.


255doesnotexist:
@guangliang2019 this pr is ready for review. the dialog component implementation is complete, including base and shadcn variants, cross-runtime body portal fix, and z-index layering. all tests are passing. please take a look when you have time.

guangliang2019 (commented):
I’ve reviewed the Preview, and it looks excellent. The code quality is also very high, and the implementation feels quite complete. I don’t expect it will take long to get this merged.

That said, this PR clearly touches on capability boundaries that Proto UI still needs to formalize. Given that you’ve previously worked on privilege-level `asHook`s, I’d like to ask for your help in filling in these missing capabilities. The current approach—gluing Portal logic directly into the DOM within the prototype layer—is not a sustainable solution.

### Proposed changes:

* Introduce Portal capability as a parameter on `asOverlay`, representing whether the mounting location should be global. This effectively enhances `asOverlay`.

  * Since this introduces a new host capability, you’ll need to extend the overlay module to request this host-cap, have each adapter implement it, and update `asOverlay` in the hooks package to use the new module capability.
  * For Web Components specifically, completely abandoning the existing DOM structure would break context propagation. A viable approach is to retain the existing structure while additionally inserting DOM at the root level.

* Another extension point is to add a “modal” parameter to `asOverlay`. A modal overlay should block interaction with the background (currently, background scrolling is still possible after opening a dialog, which indicates this capability is missing).

  * The implementation path is similar to the previous item: introduce a new host-cap via the overlay module.

* In addition, during the review, I noticed some potential issues with the dialog anatomy:

  * The cardinality of parts like `trigger`, `description`, and `close` is too restrictive. In practice, there can often be more than one.
  * The role of `overlay` is somewhat unclear. Renaming it to something like `mask` (or another more specific term) might be more appropriate. As it stands, “overlay” is hard to associate with a concrete part of dialog interaction and feels more like an implementation detail leaking to the surface.

---

255doesnotexist:
@guangliang2019 

all the structural changes from your review are in:
- overlay caps wired (`OVERLAY_GLOBAL_MOUNT_CAP`, `OVERLAY_MODAL_CAP`, `OVERLAY_LAYER_SCHEDULER_CAP`)
- `portal` / `modal` / `layerRole` added to `asOverlay` hook
- anatomy renamed `overlay` -> `mask`
- trigger/description/close cardinality max bumped to 100
- web component adapter keeps the `parentNode` override trick

but there's a runtime-level blocker that forced the adapter impl to diverge from the obvious path.

**the problem:** using react `createPortal` or vue `Teleport` inside the adapter destroys the stable host node lifecycle that `createReactHostSession` / `createVueHostSession` expects. 

when `isGlobalMounted` flips from false -> true, `Teleport` rebuilds the host DOM node. that triggers `initSession` again (`lastInitRoot` sees a new element), the old session gets disposed, and the new one starts without the overlay state synced. result: vue dialog opens but the mask/content never renders and buttons lose their text.

with react `createPortal`, the host node gets physically moved by the reconciler. this doesn't dispose the session, but it does break the layout baseline RAF timing and occasionally causes the host to unmount/remount in the middle of a transition. result: click "open dialog" and the trigger button just vanishes.

both issues only reproduce in a real browser; vitest passes because happy-dom doesn't emulate reconciler DOM moves.

**the working fix:** keep the cap interface you wanted, but implement `globalMount.mount` in react/vue adapters as direct `document.body.appendChild(hostEl)` plus `setProtoParent(hostEl, originalParent)`. then patch `web-event-router.ts` to route global `pointerdown`/`click` events back to the owning root via the `PROTO_PARENT_INSTANCE_MARK` chain.

this satisfies the design boundary (prototype layer no longer touches DOM directly) while keeping the DOM node stable for the runtime session. the event routing regression test is in `packages/adapters/web-component/test/router.spec.test.ts`.

do you think this is acceptable, or do you want the adapter layer to swallow the framework portal complexity despite the session drift? i can explore deeper reconciler workarounds if needed, but so far every `createPortal`/`Teleport` variant has broken interactively.

packages/modules/overlay/package.json
```diff
@@ -6,6 +6,7 @@
   "dependencies": {
     "@proto.ui/core": "workspace:*",
     "@proto.ui/module-base": "workspace:*",
+    "@proto.ui/module-expose-state-web": "workspace:*",
```
guangliang2019: 奇怪的依赖，overlay 是否能够不依赖 expose-state-web？那是个提供给 web 的特化拓展包，如果真的要依赖也请去依赖 expose-state 而不是 expose-state-web，可以先看看这个依赖是否有必要（暂时我认为 overlay 和 expose-state 貌似没有直接关联，asOverlay 需要用 expose + state，不代表 overlay module 需要耦合 expose-state，这里需要理由）

255doesnotexist:
@guangliang2019 replied to the review.

**portal / modal / layer scheduling**
all landed:
- `asOverlay` hook now accepts `portal`, `modal`, `layerRole`, `layerOffset`
- overlay module requests `OVERLAY_GLOBAL_MOUNT_CAP`, `OVERLAY_MODAL_CAP`, `OVERLAY_LAYER_SCHEDULER_CAP` from the host
- react / vue / web-component adapters all implement the three caps
- react/vue adapters use imperative `body.appendChild(hostEl)` in `globalMount.mount` to keep the host node stable for the runtime session (reconciler-level `createPortal`/`Teleport` destroys the session lifecycle, detailed in the earlier comment)
- web-component adapter keeps the `parentNode` override trick so context propagation still works while the node lives under body

**anatomy fixes**
- `overlay` role renamed to `mask` in `DIALOG_FAMILY`
- `trigger`, `description`, `close` cardinality max bumped from 1 to 100
- `createBodyPortal()` removed from dialog prototypes; portal behavior is now fully driven by the overlay module via caps

**dependency cleanup (your line comment)**
fixed in `0a82ed7`:
- `HOST_ELEMENT_CAP` moved from `@proto.ui/module-expose-state-web` to `@proto.ui/core`
- `module-expose-state-web` re-exports it for backward compat
- `module-overlay` now imports `HOST_ELEMENT_CAP` from `core` and the dependency on `module-expose-state-web` is dropped
- `adapter-base` wiring also imports `HOST_ELEMENT_CAP` from `core` now

all tests pass (504). anything else you want tightened before merge?

255doesnotexist:
@guangliang2019 

lockfile updated in `5c7bf46`. ci should be green now.

summary of what's on this branch since your review:
- portal / modal / layer scheduling wired through overlay caps
- anatomy renamed `overlay` -> `mask`, cardinality relaxed for trigger/description/close
- `HOST_ELEMENT_CAP` moved to `@proto.ui/core`, `module-overlay` no longer depends on `module-expose-state-web`
- react/vue `globalMount` uses imperative `body.appendChild` to keep host nodes stable (framework portal breaks session lifecycle in real browsers)
- web component keeps `parentNode` override for context propagation
- tests: 504 passing

let me know if you want anything else adjusted before merge.

guangliang2019 (approved):
非常棒，感谢你的贡献

PR #244, opened 2026-05-05
feat: add base Checkbox prototype

### LINKED ISSUE #222: Add a base Checkbox prototype
Campaign: v0-good-first-issues
Seed id: `base-checkbox-prototype`

## Summary

Add the headless base Checkbox prototype with checked, unchecked, disabled, and indeterminate coverage.

## Scope

- Add a base Checkbox prototype using the existing base prototype structure.
- Use base Switch and base Toggle as the closest references for checked/pressed boolean state and disabled behavior.
- Include an indeterminate-state example if it fits the current state model.
- Add a documentation demo and API notes.

## Acceptance criteria

- [ ] Checked and disabled states are represented through Proto UI state conventions.
- [ ] A contributor can inspect the demo source from docs.
- [ ] Existing type and runtime checks continue to pass.

## References

- `packages/prototypes/base/src/switch/root.ts`
- `packages/prototypes/base/src/switch/shared.ts`
- `packages/prototypes/base/src/switch/types.ts`
- `packages/prototypes/base/test/switch.test.ts`
- `packages/prototypes/base/src/toggle/toggle.ts`
- `packages/prototypes/base/src/toggle/types.ts`
- `packages/prototypes/base/test/toggle.test.ts`
- `apps/www/src/content/docs/en/ui-libraries/base`
- `internal/contracts/state`

## Out of scope

- Form serialization.
- Styled checkbox implementations.

## Maintainer notes

Category: `prototype`
Difficulty: `small`

shangxinxiaolin:
## Summary

Closes #222

- Add base Checkbox prototype with checked/unchecked/disabled/indeterminate states
- Checkbox root extends Toggle semantics, adding indeterminate as a custom boolean state
- Checkbox indicator reads root checked and indeterminate states via anatomy runtime access
- Press on an indeterminate checkbox clears indeterminate
- 5 unit tests covering toggle semantics, indeterminate state, disabled suppression, and indicator anatomy sync
- Docs pages (en/zh-cn) with PrototypePreviewer demos

## Test plan

- [ ] 5 unit tests pass (checkbox.test.ts)
- [ ] Type check passes
- [ ] Docs site renders Checkbox page correctly with live preview
- [ ] CI passes

### DIFF
--- a/packages/prototypes/base/src/checkbox/shared.ts
+++ b/packages/prototypes/base/src/checkbox/shared.ts
@@ -1,0 +1,9 @@
+import { createAnatomyFamily } from '@proto.ui/core';
+
+export const CHECKBOX_FAMILY = createAnatomyFamily('base-checkbox', {
+  roles: {
+    root: { cardinality: { min: 1, max: 1 } },
+    indicator: { cardinality: { min: 0, max: 1 } },
+  },
+  relations: [{ kind: 'contains', parent: 'root', child: 'indicator' }],
+});
--- a/packages/prototypes/base/src/checkbox/shared.ts
+++ b/packages/prototypes/base/src/checkbox/shared.ts
@@ -1,4 +1,11 @@
 import { createAnatomyFamily } from '@proto.ui/core';
+import type { ContextKey } from '@proto.ui/types';
+
+export type CheckboxContextValue = {
+  checked: boolean;
+  indeterminate: boolean;
+  disabled: boolean;
+};
 
 export const CHECKBOX_FAMILY = createAnatomyFamily('base-checkbox', {
   roles: {
@@ -7,3 +14,8 @@ export const CHECKBOX_FAMILY = createAnatomyFamily('base-checkbox', {
   },
   relations: [{ kind: 'contains', parent: 'root', child: 'indicator' }],
 });
+
+export const CHECKBOX_CONTEXT = {
+  __brand: 'ContextKey',
+  debugName: 'base-checkbox',
+} as ContextKey<CheckboxContextValue>;
--- a/packages/prototypes/base/src/checkbox/shared.ts
+++ b/packages/prototypes/base/src/checkbox/shared.ts
@@ -3,7 +3,7 @@ import { createAnatomyFamily } from '@proto.ui/core';
 export const CHECKBOX_FAMILY = createAnatomyFamily('base-checkbox', {
   roles: {
     root: { cardinality: { min: 1, max: 1 } },
-    indicator: { cardinality: { min: 0, max: 1 } },
+    indicator: { cardinality: { min: 1, max: 100 } },
   },
   relations: [{ kind: 'contains', parent: 'root', child: 'indicator' }],
 });
--- a/apps/www/astro.config.mjs
+++ b/apps/www/astro.config.mjs
@@ -184,6 +184,11 @@ export default defineConfig({
                   translations: { en: 'Select', 'zh-CN': 'Select' },
                   slug: 'ui-libraries/base/select',
                 },
+                {
+                  label: 'Checkbox',
+                  translations: { en: 'Checkbox', 'zh-CN': 'Checkbox' },
+                  slug: 'ui-libraries/base/checkbox',
+                },
               ],
             },
             {
--- a/apps/www/src/components/PrototypePreviewer/prototype-modules.ts
+++ b/apps/www/src/components/PrototypePreviewer/prototype-modules.ts
@@ -71,6 +71,14 @@ const manualPrototypeModules: Record<string, PrototypeModuleLoader> = {
     const mod = await import('../../../../../packages/prototypes/base/src/hover-card/content');
     registerPrototype('base-hover-card-content', mod.default);
   },
+  'base-checkbox-root': async () => {
+    const mod = await import('../../../../../packages/prototypes/base/src/checkbox/root');
+    registerPrototype('base-checkbox-root', mod.default);
+  },
+  'base-checkbox-indicator': async () => {
+    const mod = await import('../../../../../packages/prototypes/base/src/checkbox/indicator');
+    registerPrototype('base-checkbox-indicator', mod.default);
+  },
   'base-transition': async () => {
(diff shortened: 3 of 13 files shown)


packages/prototypes/base/src/checkbox/shared.ts
```diff
@@ -0,0 +1,9 @@
+import { createAnatomyFamily } from '@proto.ui/core';
+
+export const CHECKBOX_FAMILY = createAnatomyFamily('base-checkbox', {
+  roles: {
+    root: { cardinality: { min: 1, max: 1 } },
+    indicator: { cardinality: { min: 0, max: 1 } },
```
guangliang2019: 👀 Although this may sound a bit unusual, a checkbox is not necessarily limited to a single indicator. For example, one indicator may be used for the normal layout, while another may be an animated decorative element that carries the same indicator semantics.

🤔 Also, whether a checkbox should be allowed to have no indicator at all is still questionable. Can you provide an example where that would be necessary?

guangliang2019 (changes requested):
❤️ Thanks for the work here. The prototype split is very good.

👍 I especially like that Checkbox is modeled as separate `root` and `indicator` parts instead of being implemented as one large prototype. This matches Proto UI's prototype boundary direction: the indicator has its own feedback/state reflection role, so keeping it as a separate anatomy part is the right shape. Reusing `asToggle()` for the root checked behavior is also the right direction.

👀 However, I found a blocking issue in the docs preview integration.

Image

The docs page uses:

```mdx
<PrototypePreviewer demoId="demo-base-checkbox" />
````

But `demoId` expects an actual `*.demo.ts` demo module. This PR appears to add the code example map and the MDX page, but not the runnable demo definition itself. That likely explains why the preview page cannot render the live demo.

❤️ Could you add the real demo module for `demo-base-checkbox`, following the existing PrototypePreviewer demo pattern, or otherwise adjust the page to use the correct previewer API?

A few other points to check:

1. Please verify whether the Web Component tag names should be `wc-base-checkbox-root` / `wc-base-checkbox-indicator` or `base-checkbox-root` / `base-checkbox-indicator`. The tests use `base-checkbox-root`, while the docs code sample uses the `wc-` prefix.

2. Please clarify the controlled semantics of `indeterminate`. Right now `indeterminate` can be provided as a prop, but `press.commit` directly clears the internal indeterminate state and only emits `checkedChange`. If `indeterminate` is controlled, the external owner may not know it should clear `indeterminate`. We should either:

   * add a clear change event for indeterminate,
   * treat `indeterminate` as uncontrolled-only for now,
   * or document the current behavior explicitly.

3. The docs page can stay small for this PR, but it should at least mention the root / indicator split and the indeterminate behavior.

👍 Overall, the core prototype direction is strong. I don't think this needs a redesign, but the preview integration and indeterminate controlled behavior need to be tightened before merge.

guangliang2019 (changes requested):
❤️ Thanks for the PR. The overall direction looks reasonable, and the implementation quality is better than a rough draft.

👀 However, this PR is not ready to merge yet. Before the next review round, please pay attention to a few collaboration expectations:

* **1️⃣ Please always verify your changes locally before opening or updating a PR.** For GUI-related work, this also means actually opening the preview and checking the behavior by hand. It looks like you tried to fix the bug where the demo could not render correctly, but the result was probably not verified locally. This kind of workflow is generally not accepted in either open-source collaboration or regular engineering work, because it pushes basic verification work onto maintainers.

  Image

* **2️⃣ Review comments in a PR need to be resolved.** That usually means either making the requested change, or replying with a clear reason why you think the current implementation should stay unchanged. Then the maintainer can decide whether the comment can be resolved. If a PR still has unresolved review comments, it cannot be merged.

* **3️⃣ When you believe the PR is ready for another review round, please leave a comment saying so.** It is also helpful to briefly summarize what changed in this update. Otherwise, maintainers may not know whether the PR is ready to be reviewed again, and may accidentally review unfinished work. That is awkward for both sides: you may receive feedback on a half-finished state, and maintainers may spend time reviewing something that was not ready yet.

❤️ For this PR, please first address the existing review comments, run the relevant checks locally, update the test plan, and then leave a short comment requesting re-review.

shangxinxiaolin:
I addressed the latest Checkbox review feedback and verified the demo locally.

  Updates:
  - Fixed the Checkbox preview integration.
  - Synced the demo status text with the actual root state.
  - Fixed indicator state synchronization so the indicator no longer keeps the indeterminate visual state after the root
  clears it.
  - Added test coverage for indicator checked/indeterminate synchronization.

  Validation:
  - `pnpm vitest run packages/prototypes/base/test`
  - `pnpm --filter apps-www check`
  - Manually checked `/zh-cn/ui-libraries/base/checkbox/` locally:
    - the live demo renders without `PrototypePreviewer` errors
    - `Indeterminate` changes to `checked: true, indeterminate: false` after click
    - `Disabled` remains unchanged
    - Console only shows the existing `StarlightThemeProvider is not defined` site-level error, with no Checkbox/
  PrototypePreviewer errors

  Ready for re-review.

guangliang2019 (changes requested):
👍 Basically no major issues remain. I expect this can be merged after the next update.

❗️ The remaining issue is that the demo styling is ambiguous. A checkbox is generally not recommended to be fully circular; it should preferably be square, possibly with rounded corners. A fully circular shape can easily be confused with a radio button, so this still needs to be changed.

Image

shangxinxiaolin:
Updated.

  - Changed the checkbox demo indicator from theme `rounded` to explicit `rounded-[4px]`, so it now reads as a
  square checkbox instead of a circular/radio-like control.
  - Updated both the live demo config and WC/React/Vue code snippets.
  - Verified with `pnpm --filter apps-www check`.

  Commit: 7741bdbf

guangliang2019 (approved):
🎉🎉🎉 Congratulations, this PR has reached merge quality. Great work 👍.

guangliang2019:
Given that prototypes are not exactly easy to write, thank again you very much for the contribution. ❤️

🧭 If you are interested, you can try working on the shadcn version of Checkbox next. It would directly depend on the `base-checkbox` prototypes you just contributed, and use them to implement a component with a more modern design style.

PR #245, opened 2026-05-06
feat(whitepaper): add Mermaid information-flow diagram (#234)

### LINKED ISSUE #234: [docs] Add an information-flow diagram to the whitepaper
## Goal

Add one clear, reusable diagram to the English and Chinese **Information Flow Model** whitepaper pages. The diagram should help a reader understand the relationships without introducing or changing Proto UI semantics.

This is a documentation projection task. The contributor is not expected to design a new protocol or make contract decisions.

## Current state

The two whitepaper pages explain the model in prose but do not yet contain a diagram:

- `apps/www/src/content/docs/en/whitepaper/information-flow-model.md`
- `apps/www/src/content/docs/zh-cn/whitepaper/information-flow-model.md`

The current catalog contains the semantic context below. These entities are currently `draft`, so treat them as the cataloged direction and report contradictions instead of silently resolving them:

- `spec/knowledge/K-INFORMATION-CHANNEL-0001.yaml`
- `spec/knowledge/K-COMPONENT-ACTOR-0001.yaml`
- `spec/knowledge/K-COMPONENT-INTERACTION-0001.yaml`
- `spec/contracts/C-CORE-CHANNEL-0001.yaml`

Use the channel contracts to verify labels and direction:

- `spec/contracts/C-EVENT-0001.yaml`
- `spec/contracts/C-FEEDBACK-0001.yaml`
- `spec/contracts/C-PROPS-0001.yaml`
- `spec/contracts/C-EXPOSE-0001.yaml`
- `spec/contracts/C-CONTEXT-0001.yaml`
- `spec/contracts/C-STATE-0001.yaml`

## Suggested workflow

1. Read the two whitepaper pages and the catalog entities above.
2. Propose a compact visual hierarchy for the component, its interaction targets, the directional channels, and the internal dimensions described by the article.
3. Reuse one visual implementation where practical; localize surrounding explanation and accessible text.
4. If the prose and catalog disagree, call out the mismatch in the PR instead of inventing a resolution.
5. Verify the site in both light and dark themes and at narrow viewport width.

AI-assisted exploration and implementation are welcome. A useful AI handoff should include the source paths consulted, the chosen visual mapping, and any ambiguity that still needs maintainer review.

## Acceptance criteria

- [ ] The diagram appears in both the English and Chinese Information Flow Model pages.
- [ ] Labels and arrow directions match the cited whitepaper/spec sources.
- [ ] The visual introduces no new actor, channel, lifecycle, or protocol claim.
- [ ] The diagram remains readable on narrow screens and in both light and dark themes.
- [ ] Equivalent alt text or nearby prose makes the information understandable without the image.
- [ ] The implementation avoids duplicating the full visual structure for each locale unless localization requires it.
- [ ] `corepack pnpm@10.32.1 --filter apps-www build` passes.

## Out of scope

- Rewriting the Information Flow Model.
- Changing any `K-*`, `C-*`, or other spec entity.
- Adding animation.
- Redesigning unrelated whitepaper pages.
- Introducing a new site-wide diagram framework.

## Maintainer notes

This remains a Good First Issue because the semantic sources and verification boundary are provided. The contribution should stay within visual/documentation projection; any semantic decision returns to maintainers.

shangxinxiaolin:
## Summary

在白皮书信息流模型页面添加 Mermaid 示意图，展示 intent、props、state、events、feedback、host rendering 之间的连接关系。

- en/zh-cn 两页均添加
- 标签与白皮书术语一致
- 包含 alt text 说明文字
- Closes #234

### DIFF
--- a/apps/www/src/components/override/ThemeProvider.astro
+++ b/apps/www/src/components/override/ThemeProvider.astro
@@ -21,6 +21,19 @@
       // 给 chrome/safari 一个暗色介面提示（和 meta 协同）
       document.documentElement.style.colorScheme = t;
     };
+    const updatePickers = (theme = load() || 'auto') => {
+      document.querySelectorAll('starlight-theme-select').forEach((picker) => {
+        const select = picker.querySelector('select');
+        if (select) select.value = theme;
+
+        const tmpl = document.getElementById('starlight-theme-icons');
+        const newIcon = tmpl && tmpl.content.querySelector('.' + theme);
+        const oldIcon = picker.querySelector('svg.label-icon');
+        if (newIcon && oldIcon) {
+          oldIcon.replaceChildren(...newIcon.cloneNode(true).childNodes);
+        }
+      });
+    };
 
     // 初始化（若无存储，则跟随系统）
     const initial = load() || getSystem();
@@ -48,6 +61,12 @@
       hydratePickers() {
         document.dispatchEvent(new CustomEvent('starlight-theme:hydrate'));
       },
+    };
+
+    // Compatibility for Starlight's built-in ThemeSelect used in the default
+    // mobile sidebar preferences slot.
+    window.StarlightThemeProvider = {
+      updatePickers,
     };
 
     // 如果恰好监听到了主题切换，那么随系统切换主题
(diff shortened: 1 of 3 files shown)


shangxinxiaolin:
Updated again to restore compatibility with Starlight's built-in theme selector in the mobile sidebar.

  Verified locally:
  - `corepack pnpm@10.0.0 --filter apps-www check`
  - `corepack pnpm@10.0.0 --filter apps-www build`
  - Browser console on the whitepaper page is clean now

  Ready for re-review.

guangliang2019 (changes requested):
❤️ Thanks for working on this.

❌ After reviewing the direction, I think this PR should be reworked rather than merged in its current form.

⚠️ The main issue is that Mermaid diagrams are not a good fit for this part of the documentation. For these conceptual sections, we should use hand-drawn-style illustrations instead, for example illustrations made with Excalidraw.

🍵 The reason is not just visual style. When a diagram needs Mermaid to express the idea, it usually means the logic has already become too complex or too procedural. That is exactly what we want to avoid here.

For this part of Proto UI, the diagrams should be:

- suggestive rather than exhaustive
- helpful for understanding rather than formally precise
- closer to conceptual illustrations than flowcharts
- simple enough that they can be understood at a glance

❗️ In other words, we need “illustrations”, not “flowcharts”. These diagrams should support the reader’s intuition, not try to precisely encode the philosophy of the project.

💡 I will attach some illustrations used in the talk below as references for the expected direction.

Image

⚠️ Also, for this task specifically, I want to **reject AI-generated** diagram output. The diagrams need to be intentionally designed by a human, because the goal is not just to produce a valid visual representation, but to decide what should be emphasized, simplified, or omitted.

Issue #224, opened 2026-05-04
Add a base Tooltip prototype


guangliang2019:
Campaign: v0-good-first-issues
Seed id: `base-tooltip-prototype`

## Summary

Add a base Tooltip prototype that demonstrates trigger, content, open state, and delay configuration.

## Scope

- Use existing overlay and trigger concepts where possible; base Hover Card is the primary anatomy and open-state reference.
- Add a minimal docs demo with hover/focus open behavior.
- Document any behavior intentionally left to adapters or host runtimes.

## Acceptance criteria

- [ ] The prototype has clear trigger and content anatomy.
- [ ] Open state is observable in the demo.
- [ ] The issue does not require a positioning engine.

## References

- `packages/prototypes/base/src/hover-card/root.ts`
- `packages/prototypes/base/src/hover-card/trigger.ts`
- `packages/prototypes/base/src/hover-card/content.ts`
- `packages/prototypes/base/src/hover-card/types.ts`
- `packages/prototypes/base/test/hover-card.test.ts`
- `packages/prototypes/base/src/dialog/trigger.ts`
- `packages/prototypes/base/src/dialog/content.ts`
- `packages/prototypes/base/test/dialog.test.ts`
- `internal/contracts/overlay/as-overlay.v0.md`
- `packages/modules/as-trigger`

## Out of scope

- Collision-aware floating positioning.
- Rich hover-card behavior.

## Maintainer notes

Category: `prototype`
Difficulty: `small`

guangliang2019:
Closing as not planned in its current form.

Tooltip 涉及 hover/focus intent、delay、overlay/positioning、dismissal 与跨 adapter 行为，原 Issue 明确排除 positioning 后已经无法代表一个完整可承诺的协议。

The original v0 campaign treated Base Tooltip as a small implementation task. Under the current spec-first workflow, a new official prototype needs an agreed prototype identity, criteria, relationships, executable coverage, and documentation projection before implementation can be safely delegated. That semantic work is maintainer-owned and is not a Good First Issue.

If Base Tooltip becomes a current priority, we will open a new evidence-backed proposal with the contract boundary already narrowed, then publish only the remaining contributor-safe slice.

PR #248, opened 2026-05-06
feat: add base Tooltip and Slider prototypes

shangxinxiaolin:
## Summary
- Add base Tooltip prototype (#224): trigger/content anatomy, hover/focus open, controlled/uncontrolled, delay config - e5bb67a`n- Add base Slider prototype (#223): value/min/max/step/disabled/orientation, slide.commit event, controlled/uncontrolled - 38bfd7f1`n
Closes #224, closes #223

PR #250, opened 2026-05-07
fix(base): complete Tooltip prototype

shangxinxiaolin:
## Summary

  - Add/fix the base Tooltip prototype as a standalone PR for #224.
  - Wire Tooltip into the docs previewer so the demo can actually load.
  - Make `delay` participate in hover/focus opening behavior.
  - Add an observable open-state label in the docs demo.
  - Document that positioning/collision/portal details are left to host runtimes or adapters.

  ## Validation

  - [x] `pnpm vitest run packages/prototypes/base/test/tooltip.test.ts packages/prototypes/base/test/hover-card.test.ts
  packages/prototypes/base/test/dialog.test.ts`
  - [x] `pnpm --filter apps-www check`
  - [x] Manual visual check on the Tooltip docs demo

  ## Notes

  This PR intentionally keeps Tooltip separate from the other prototype work. It does not include the local `radio` worktree
  changes.


shangxinxiaolin:
I closed the previous mixed PR #248. This PR keeps Tooltip isolated from Slider as requested.

  I also updated the Tooltip implementation to fix the open-state behavior:
  - `delay` now applies to hover/focus opening
  - disabled Tooltip roots force the tooltip closed
  - `defaultOpen` is honored
  - content now follows the root `open` state as the source of truth
  - the docs demo shows the observable open state

  Validation:
  - `pnpm vitest run packages/prototypes/base/test/tooltip.test.ts packages/prototypes/base/test/hover-card.test.ts
  packages/prototypes/base/test/dialog.test.ts`
  - `pnpm --filter apps-www check`
  - Manual docs demo visual check

shangxinxiaolin:
Updated again.

  - Resolved the `prototype-modules.ts` merge conflict by keeping both tooltip and checkbox prototype loaders.
  - Fixed the local runtime error from `StarlightThemeProvider` so the tooltip docs page no longer reports
  console errors.
  - Verified locally:
    - `pnpm --filter apps-www check`
    - `pnpm --filter apps-www build`
    - manually opened `/zh-cn/ui-libraries/base/tooltip/` and confirmed the tooltip demo works without console
  errors.

  Ready for re-review.

guangliang2019 (changes requested):
❤️ Thanks for working on this. The demo wiring, docs, and tests are useful, and I also appreciate the ThemeProvider fix — that stale Starlight theme issue has been around for a while.

👀 However, I don’t think the current Tooltip implementation is ready to merge. The main issue is not code style, but that the current design simplifies Tooltip semantics too much and crosses some Proto UI abstraction boundaries.

## ThemeProvider

👍 I’m okay with the ThemeProvider fix itself. But process-wise, it should be split out from this Tooltip PR.

🙏 Please create a separate issue for the stale Starlight ThemeProvider problem and move that fix into its own PR. I can review and approve it independently.

## Tooltip needs a redesign

❌ The current Tooltip implementation should be reworked around the correct protocol-level semantics.

### 1. Please revisit the Anatomy first

💡 Please use Radix UI / Base UI as references for Tooltip anatomy, but don’t copy them mechanically.

👀 Proto UI’s anatomy should follow Proto UI’s own whitepaper and existing prototypes first. For example, in Proto UI, Portal is currently an option/capability of `asOverlay`, not a standalone prototype part. So our anatomy may differ from common Headless UI libraries.

🤔 A rough direction to consider:

- `TooltipGroup` — probably better than `TooltipProvider` as the public prototype name. Technically it works like a provider, but semantically it represents a group of related tooltips. Proto UI should prioritize interaction semantics over implementation naming.
- `TooltipRoot`
- `TooltipTrigger`
- `TooltipOverlay`
- `TooltipContent`
- `TooltipArrow`(or `TooltipIndicator`)
- ...

🤕 This should also be compared against the existing Hover Card implementation, because Tooltip is structurally similar to Hover Card, but more complex due to delay behavior and group-level coordination.

### 2. Tooltip content is overlay content

👀 Tooltip content should be modeled as overlay content. The current implementation/demo relies too much on local CSS classes like `absolute left-0 top-full` for placement. That is not the right semantic model for a base Tooltip prototype.

⚠️ It is fine if collision handling, arrows, or portal behavior are not fully completed in this PR. But the base prototype should not encode a misleading positioning model. It should build on the existing overlay abstraction.

### 3. Do not bypass the Proto UI protocol through host DOM events

⚠️ The current delayed-open logic dispatches a host DOM event and receives it back through `host.tooltip-delay-open`. That is not acceptable for an official base prototype.

👀 Official prototypes should not bypass the protocol by binding themselves to a concrete host. If Tooltip needs delayed interaction behavior, we should either use an existing protocol-level abstraction or introduce a small new one.

💡 The transition-related work may provide useful prior art here, since it already handled delayed unmounting, which is more complex than a simple delayed open.

### 4. Delay may reveal a missing abstraction

🤔 The delay behavior is probably the most important missing piece here.

⚠️ Please do not use `setTimeout`, `queueMicrotask`, Promise microtasks, or host-specific event bridges directly inside the official prototype. That would implicitly assume a Web-like host and thread model.

👍 If Proto UI currently lacks the required timing abstraction, we should introduce a small dedicated abstraction first. Even a temporary one is acceptable, but it should be centralized and easy to replace later. In other words, please don’t scatter direct timer logic through the Tooltip implementation.

### 5. Tooltip Trigger should not compose `asButton()`

⚠️ Tooltip Trigger is not necessarily a button. A tooltip can be attached to text, icons, links, inputs, or arbitrary interactive elements.

👀 This is different from Dialog Trigger or Dialog Close, where button semantics are much more reasonable. Tooltip Trigger is closer to Hover Card Trigger: it should provide tooltip-related hover/focus tracking, not force button semantics onto the target.

### 6. TooltipGroup likely needs Focusable / FocusGroup

👍 Tooltip will likely need `Focusable` and `FocusGroup`, especially when implementing group-level behavior.

👀 One important TooltipGroup behavior is: when one tooltip in the group is already open, moving to another tooltip trigger should usually skip the normal delay and show the next tooltip immediately.

🍵 This behavior is easy to miss, but it is part of the interaction model. The current root-local timing model does not leave enough room for it.

### 7. State-machine details can be reviewed after the redesign

👀 There may also be smaller issues around `defaultOpen`, `disabled`, and controlled/uncontrolled state synchronization. But I don’t think it is useful to review those details deeply before the larger structural issues are fixed, because a redesign may change the state model substantially.

## Demo / self-check

⚠️ One more practical issue: the Tooltip docs page appears to be implemented, but it is not added to the website sidebar. Because of that, I could not access the real docs page from the site navigation and could not fully preview whether the actual demo renders correctly.

🙏 Please make sure the docs page is linked from the sidebar, and please do a real self-check before requesting review again:

- open the docs page from the sidebar/navigation;
- verify that the demo renders correctly;
- verify hover/focus behavior manually;
- verify runtime switching if the previewer supports multiple runtimes for this demo;
- confirm the exact validation commands you ran.

## Summary

❤️ So I’m requesting changes for now.

💡 This PR should not be patched forward as a simplified host-oriented Tooltip. It needs to be redesigned around:

- Overlay
- Focusable / FocusGroup
- Hover Card as the closest existing structural reference
- TooltipGroup-level delay coordination
- a protocol-level timing/delay abstraction, if one is currently missing
- a real docs/demo preview path through the website sidebar

❤️ Thanks again for pushing this forward. Tooltip is a good component to work on, but it touches several important Proto UI boundaries, so we need to be stricter here.

shangxinxiaolin:
Thanks for the detailed review! Here's a summary of what has been addressed:

ThemeProvider

🙏 Please create a separate issue for the stale Starlight ThemeProvider problem and move that fix into its own PR.

⏳ Acknowledged. Will split this into a separate issue + PR. Not included in this round of changes.

1. Revisit the Anatomy

💡 Please use Radix UI / Base UI as references for Tooltip anatomy.

✅ Done. Studied the existing HoverCard and Dialog implementations as structural references. The anatomy has been expanded from 3 roles to 5 roles within the family:

TooltipGroup (separate prototype, optional)
TooltipRoot
  ├── TooltipTrigger
  └── TooltipOverlay       ← NEW: uses asOverlay for positioning/visibility
        ├── TooltipContent  ← CHANGED: now a pure semantic content wrapper
        └── TooltipArrow    ← NEW: decorative arrow indicator
The overlay/content split follows the same principle as Dialog (where mask and content are separate overlay-aware parts). Portal remains a capability of asOverlay, not a standalone part, consistent with Proto UI's existing model.

2. Tooltip content is overlay content

⚠️ It should build on the existing overlay abstraction.

✅ Done. TooltipOverlay uses asOverlay with placement: 'top' and sideOffset: 4. Demo code no longer uses manual positioning CSS like absolute left-0 top-full. The new structure clearly separates overlay mechanics (in overlay.ts) from content semantics (in content.ts).

3. Do not bypass the Proto UI protocol through host DOM events

⚠️ The current delayed-open logic dispatches a host DOM event.

✅ Done. No host DOM events (host.tooltip-delay-open or similar) are used. Delay is handled via the asDelay() hook, which is a protocol-level abstraction.

4. Delay may reveal a missing abstraction

⚠️ Please do not use setTimeout, queueMicrotask, Promise microtasks directly.

✅ Done. The implementation uses asDelay() from @proto.ui/hooks. No direct setTimeout, queueMicrotask, or Promise microtask usage in the prototype code.

5. Tooltip Trigger should not compose asButton()

⚠️ Tooltip Trigger is not necessarily a button.

✅ Done. TooltipTrigger uses asFocusable() — no asButton composition. TooltipTriggerExposes no longer includes pressed or click.

6. TooltipGroup likely needs Focusable / FocusGroup

👀 When one tooltip in the group is already open, moving to another tooltip trigger should skip the normal delay.

✅ Done. TooltipGroup uses asFocusGroup and provides a group context with openCount and skipDelayDuration. Each TooltipRoot subscribes to the group context (via trySubscribe) and skips the configured delay when openCount > 0. Test verifies this behavior: second tooltip opens immediately when the first is already open.

7. State-machine details

👀 Can be reviewed after the redesign.

Acknowledged. The redesign is now in place — happy to iterate on defaultOpen / disabled / controlled state details in a follow-up review.

Demo / self-check

🙏 Please make sure the docs page is linked from the sidebar.

✅ Done. Self-check results:

Sidebar entry exists in astro.config.mjs (slug: ui-libraries/base/tooltip)
Page returns HTTP 200 at /zh-cn/ui-libraries/base/tooltip/
Demo code updated for all 3 runtimes (wc / react / vue) with the new anatomy
All 9 tooltip tests pass, full suite 592/592 pass
Validation command: npx vitest run --reporter=verbose packages/prototypes/base/test/tooltip.test.ts

guangliang2019 (changes requested):
❤️ Thanks for the update. This round is much closer to the intended direction.

👍 The improvements are clear:

- Tooltip is now linked from the sidebar.
- The anatomy is closer to the intended model: `TooltipGroup / Root / Trigger / Overlay / Content / Arrow`.
- `TooltipTrigger` no longer composes `asButton()`.
- Tooltip visibility/positioning moved toward `Overlay + asOverlay`.
- The previous host DOM event bridge was removed.
- There is now a group skip-delay test.

🍵 That said, I still need to request changes. The remaining issues are mostly about protocol boundaries and the overlay baseline behavior.

## 😭 1. `asDelay` is the wrong abstraction shape

🗑️ Please remove `asDelay` from `@proto.ui/hooks`.

🤔 Even if Proto UI needs a timing abstraction, it probably should not be an `asHook`. It is more likely to be a core-level syntax or a special module that every adapter/runtime needs to wire, similar to a cross-platform `Time` capability. Such a capability would include delay, clock, timer, and possibly related scheduling behavior.

❌ So the current `asDelay` has the wrong public position. It turns a temporary workaround into a public hook API.

⚠️ For this PR, please do **not** modify core yet. Instead, avoid introducing a public hook and assume that a core-level `Time` capability will exist later.

A reasonable temporary direction:

- keep the timing wrapper internal to Tooltip;
- name it clearly as a temporary/debt abstraction;
- leave an obvious debt marker in the code;
- keep the call surface small so it can later be replaced by a real core/runtime `Time` capability with minimal changes.

⚠️ The important point is: do not expose `asDelay` from `@proto.ui/hooks`, and do not make this look like the final protocol-level answer.

## 💡 2. `asOverlay` should provide minimal absolute positioning feedback

👍 The new demo uses the overlay semantics, which is good. But the overlay currently has no baseline positioning feedback, so it does not visually behave like a floating layer.

💡 This exposes an existing Proto UI issue: `asOverlay` probably needs a minimal default feedback style.

🤔 In non-portal mode, the overlay follows normal parent layout unless the prototype author adds extra `relative/absolute` styles. In portal mode, it behaves more like a positioned layer. These two modes should not diverge that much at the basic overlay level.

🙏 Please update `asOverlay` so it contributes a minimal default style:

```ts
def.feedback.style.use(tw('absolute'))
````

or the equivalent style helper used in that file.

⚠️ We do **not** need to decide here which element owns `relative` positioning. That can remain the responsibility of the maker or the higher-level prototype library author. But the overlay itself should at least opt into absolute positioning as its minimum floating-layer behavior.

🍵 After that, TooltipOverlay can keep its visual styling, but it should not need to manually rediscover the basic overlay positioning baseline.

## 👀 3. Tooltip demo should visually verify overlay behavior

📒 The current demo should verify:

* the tooltip looks like a floating layer, not normal inline content;
* hover/focus opens and closes it correctly;
* runtime switching still works;
* the group skip-delay behavior is either shown in the demo or documented with a clear example.

## 🙏 4. ThemeProvider is still unrelated to this PR

✨ I still appreciate the ThemeProvider fix, but it is unrelated to Tooltip.

❤️ Please split it into a separate issue/PR if possible. At minimum, please avoid mixing unrelated site fixes into component prototype PRs after this one. Clean PR scope matters for review and regression tracking.

## 🍵5. Smaller state details can follow the structural fixes

💡 There may still be smaller state-machine issues around `disabled`, controlled/uncontrolled open state, and group open-count synchronization. But I think the highest-priority fixes are now:

1. remove public `asDelay`;
2. leave a clearly marked temporary Tooltip-local timing abstraction;
3. add the minimal `absolute` feedback to `asOverlay`;
4. re-check the real docs demo visually.

❤️ Once those are fixed, I can do another pass on the remaining state details.

cyjin-yl:
Current-status review: this implementation has been superseded architecturally by #352, which starts from current `main` and now has maintainer approval. The successor preserves the useful controlled/uncontrolled Tooltip request direction while using the current Context/Delay/Overlay/Transition/A11y capabilities, and adds Tooltip Group coordination plus additive `aria-describedby` ownership. The remaining file-only differences here (`as-delay`, old `.ts` Tooltip parts, and the older Arrow/Overlay split) belong to the pre-catalog architecture and should not be conflict-resolved into current `main`. Recommendation: close this PR as superseded after #352 merges; any still-useful visual/Arrow work should be proposed separately against the approved Tooltip protocol.

PR #256, opened 2026-05-12
feat: decouple generated styles from Tailwind and harden local pnpm usage

guangliang2019:
**Description**

This PR completes the CLI acceptance fixes from this round of work.

## Changes

- Fixed generated component export names.
  - React host files now export names like `ShadcnButton`.
  - Root component index now aliases them as `ReactShadcnButton`.
  - Web Component exports were kept aligned with the same registry-driven naming path.

- Reworked generated style output to avoid Tailwind project assumptions.
  - Prototype style tokens are now emitted through `data-pui-style` instead of adapter-owned `class` tokens.
  - Generated CSS targets `[data-pui-style~="..."]` selectors directly.
  - Removed reliance on Tailwind-only behavior such as project content scanning, `@theme`, and `@source`.
  - Renamed the generated style entry to `proto-ui-style.css`.

- Added prefixed Proto UI theme variables.
  - Theme variables are emitted as `--pui-*`.
  - Generated token CSS references `--pui-*` variables.
  - User-facing utility-style token names remain unchanged.

- Preserved user classes across adapters.
  - React, Vue, and WebComponent adapters now keep user `className` / `class` separate from Proto UI-owned style tokens.
  - User classes remain available for overrides.

- Fixed Vue adapter typing.
  - Relaxed Vue runtime typing so `createVueAdapter(Vue)` works with the Vue module import shape.

- Updated website/docs consumption path.
  - The docs app now consumes the generated `proto-ui-style.css` flow.
  - The website continues to serve as an integration verification target.

- Hardened local package manager usage.
  - Added pnpm version guard config.
  - Pinned local scripts, docs scripts, website scripts, and Husky pre-commit to `corepack pnpm@10.32.1`.
  - Updated README install commands so contributors are not blocked by a globally installed older pnpm.

## Verification

- Ran CLI init / generated-style tests.
- Ran adapter style-token tests for React, Vue, and WebComponent.
- Ran workspace runtime tests.
- Verified `corepack pnpm@10.32.1 install --frozen-lockfile` works.
- Verified old global `pnpm` fails fast with a clear version error instead of entering repeated module purge prompts.


cyjin-yl (changes requested):
gemini found real bugs but misdiagnosed the ci failure. the tailwind removal is intentional per the pr description and should not be treated as a regression.

**1. web component `data-pui-style` overwrite — confirmed bug.**

`createOwnedTwTokenApplier` replaces the entire `data-pui-style` attribute instead of merging adapter-owned tokens with pre-existing ones. the old classList logic selectively added/removed only owned classes, allowing multiple sources to coexist. now, if base styles and feedback styles both target the same host, the last writer wins and transiently wipes the other.

`applyFeedbackStyleTokensToHost` does save/restore the previous attribute value, but during active feedback the base styles are lost. with classList, base and feedback tokens coexisted; now they do not.

**2. ci `cli-smoke` failure — not a vue type error.**

`type-check` is green. the actual failure is `.github/workflows/ci.yml` still grepping for old export names (`export const Button = adapt(shadcnButton)`, `Button as ReactButton`) after the pr renamed everything to `ShadcnButton`. the pr updated `cli.test.ts` but missed the workflow.

**3. tailwindcss command — intentional decoupling, but the alias is too soft.**

re-reading the pr body: "Removed reliance on Tailwind-only behavior." the goal is correct. however, silently aliasing `tailwindcss` to `style` is a footgun. a user running `proto-ui tailwindcss --out ./src/styles/tailwindcss.css` will get a proto-ui entry file that does not import tailwindcss, without any warning.

in a migration pr, do **not** restore backward compatibility. instead, make the old command fail hard with a clear message: `the tailwindcss command has been removed. use proto-ui style instead.` then route `style` normally.

**4. vue adapter unused variable — minor hygiene.**

`const { class: className, ... }` leaves `className` unread. not a ci blocker, but worth prefixing with `_class`.

**recommended fixes before merge:**
- merge `data-pui-style` tokens instead of overwriting in `createOwnedTwTokenApplier` (and ideally `applyFeedbackStyleTokensToHost` during active state).
- update `.github/workflows/ci.yml` grep assertions for `ShadcnButton` / `ReactShadcnButton` / `VueShadcnButton` / `ShadcnButtonElement`.
- replace the soft `tailwindcss` alias with a hard error that points users to `style`.
- optionally prefix the unused `className` with `_class`.

cyjin-yl:
gemini found real bugs but whiffed the root cause of the red ci.

`data-pui-style` overwrites in `feedback-style.ts` are genuine regressions. the old classList merge preserved foreign classes; the new attribute logic does not. this will break any host that carries both base and feedback tokens simultaneously.

the `tailwindcss` command aliasing to `style` is **intentional** per the pr description ("removed reliance on tailwind-only behavior"). my previous reply misread this as a silent break. it is not.

that said, in a migration pr, a soft alias is the wrong cut. do not restore backward compatibility. instead, `proto-ui tailwindcss` should throw a hard error with a clear pointer to `proto-ui style`. silently emitting the wrong css format under an old command name is worse than breaking loudly.

as for the ci failure: it is not an unused local in the vue adapter. `type-check` is green. the smoke job dies because `.github/workflows/ci.yml` still greps for `export const Button = adapt(shadcnButton)` after you renamed everything to `ShadcnButton`. you updated the unit tests but forgot the workflow.

fix the merge logic, fix the ci greps, and make the old command die loudly. then this ships.

guangliang2019 (commented):
Thanks for the work here. I think the overall direction is right.

Decoupling the generated style output from Tailwind while keeping the style-token shape Tailwind-like is a good direction. It solves several problems at once: generated components no longer depend on the user's Tailwind content scanning, Proto UI-owned style tokens no longer compete with user `class` / `className`, and user classes can now be preserved as real user overrides.

That said, I do not think this is ready to merge yet. There are a few issues that should be fixed before merge.

## 1. `data-pui-style` needs ownership-aware merging internally

I agree that `data-pui-style` should not become a public extension surface for app makers. App makers should keep using `className` / `class` / `hostClassName` / `hostStyle` for their own styling.

The problem is not mainly whether users can manually write new `data-pui-style` tokens. The stronger issue is that base styles, feedback styles, rule-driven styles, and transient state styles may all target the same host. The current implementation makes the attribute last-writer-wins, while the previous classList path allowed independently owned token sets to coexist.

For example, `createOwnedTwTokenApplier` now writes the entire `data-pui-style` attribute. `applyFeedbackStyleTokensToHost` also temporarily replaces the whole attribute value. This can wipe tokens produced by another internal style source during the active state.

Please change this so each internal token source only owns and updates its own token set:

- read the current `data-pui-style` tokens,
- remove only the tokens previously owned by that applier/source,
- preserve other tokens,
- merge in the next owned tokens,
- and clear only the owned tokens.

Please also add regression coverage for this. A useful case would be: base tokens and feedback tokens both apply to the same host, and applying/clearing one source does not remove the other source's tokens.

## 2. Update the CI smoke assertions for the new generated names

The new generated names look correct to me, but the workflow still appears to check the old generated output names.

The tests were updated to expect names like:

```ts
export const ShadcnButton = adapt(shadcnButton);
export { ShadcnButton as ReactShadcnButton } from './react';
````

but the CI smoke assertions still need to be updated accordingly. Please update the workflow checks for the new React / Vue / Web Component generated names.

## 3. The legacy `tailwindcss` command should fail loudly instead of aliasing to `style`

The Tailwind decoupling is intentional, so I do not think we should restore the old Tailwind-specific behavior.

However, silently aliasing `proto-ui tailwindcss` to `proto-ui style` is risky. A user running an old command named `tailwindcss` may reasonably expect Tailwind-specific output, but the new output is a Proto UI style entry file.

Since this is a breaking migration, I prefer a hard error with a clear message, for example:

```txt
The tailwindcss command has been removed. Use `proto-ui style` instead.
```

That is safer than silently producing a different CSS format under the old command name.

## 4. Vue runtime typing: valid fix target, but the current loosening is a bit too broad

The Vue typing change makes sense as a fix target. The CLI-generated code currently uses:

```ts
import * as Vue from 'vue';

const adapt = createVueAdapter(Vue);
```

and the previous `VueRuntime.watch` type was too narrow to accept Vue's overloaded `watch` function. So I agree this PR should make `createVueAdapter(Vue)` type-check.

That said, the current fix weakens the runtime type quite a lot:

```ts
watch: (...args: any[]) => any;
nextTick: (...args: any[]) => Promise<any>;
ref: <T>(v: T) => { value: any };
shallowRef: <T>(v: T) => { value: any };
```

This is acceptable as a pragmatic unblocker, but I would prefer a narrower shape that still accepts the real Vue module import. For example, keep `ref` / `shallowRef` generic over `{ value: T }`, and only loosen `watch` enough to handle Vue's overloads:

```ts
watch: (
  source: any,
  cb: (...args: any[]) => void | Promise<void>,
  options?: any
) => unknown;
nextTick: (fn?: () => void) => Promise<void>;
```

Alternatively, since this is the Vue adapter, we can consider typing `VueRuntime` directly from Vue's exported function types.

I do not think this should block the PR by itself, but the current `any` shape should be treated as a compatibility fix rather than the final adapter typing.

## 5. CLI-owned utility rendering needs stronger follow-up coverage

This PR effectively introduces a Proto UI-owned renderer for a Tailwind-like utility subset. I am fine with that direction, but it means the CLI is now responsible for supporting every style token emitted by the official prototype libraries.

For this PR, I would like at least enough coverage to ensure the current generated base/shadcn token set does not silently lose styles. Unsupported tokens should not quietly disappear into comments without some kind of test or warning strategy.

This can be expanded in a follow-up, but we should not leave the new renderer without a clear safety net.

---

So my position is:

* The overall design direction is approved.
* The `data-pui-style` ownership issue, CI smoke mismatch, and legacy command behavior should be fixed before merge.
* The Vue typing and utility-renderer coverage can be improved either in this PR or as immediate follow-up work, but they should not be ignored.

guangliang2019:
@255doesnotexist Thanks for the detailed review. I addressed the requested follow-ups in `bb3de687`.

Summary of changes:

- Made `data-pui-style` ownership-aware internally.
  - Each internal style source now only removes/replaces the tokens it owns.
  - Base/setup tokens, feedback tokens, and transient owned tokens can coexist on the same host.
  - Added regression coverage for applying and clearing one source without wiping another.

- Updated the CI smoke assertions for the new generated names.
  - React now checks `ShadcnButton` / `ReactShadcnButton`.
  - Vue now checks `ShadcnButton` / `VueShadcnButton`.
  - Web Component now checks `ShadcnButtonElement`.

- Changed the legacy `proto-ui tailwindcss` command to fail loudly.
  - It now errors with: `The tailwindcss command has been removed. Use proto-ui style instead.`
  - This avoids silently producing the new Proto UI CSS format under the old Tailwind-specific command name.

- Narrowed the Vue runtime compatibility typing.
  - `ref` / `shallowRef` now preserve `{ value: T }`.
  - `watch` is only loosened enough to accept Vue's overloaded `watch`.
  - `nextTick` is typed as `(fn?: () => void) => Promise<void>`.

- Added CLI renderer coverage for the official prototype token set.
  - The test now verifies the generated base/shadcn token CSS does not contain unsupported-token comments.

Verification run locally:

```sh
corepack pnpm@10.32.1 exec vitest run packages/adapters/web-component/test/feedback-style.test.ts packages/cli/test/cli.test.ts packages/adapters/vue/test/framework-contract.test.ts
corepack pnpm@10.32.1 -s check:types:workspace
corepack pnpm@10.32.1 exec prettier --check ...
```

baiyucore (approved):

cyjin-yl:
### Code review

Found 1 issue:

1. `applyFeedbackStyleTokensToHost` switched from the `addedByAdapter` model (stack-safe — only removes the tokens this call added) to whole-attribute snapshot + restore. Two overlapping feedbacks on one host now clobber each other: the second `apply` overwrites the first's tokens (`setStyleAttribute` is unconditional), and any undo restores a stale `previous` snapshot. This was correct pre-PR — `addedByAdapter` made interleaved hover/focus/pressed effects on a single host stack-safe. Note that this helper is not re-exported from `packages/adapters/web-component/src/index.ts` (which only re-exports `adapt`/`commit`/`props`), so there is no in-tree caller today; the regression is latent in the public-looking API surface rather than active.

https://github.com/Proto-UI/Proto-UI/blob/3a80403edd1ef4f0f8e1f78eb98f3e9566b565fd/packages/adapters/web-component/src/feedback-style.ts#L14-L24

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #329, opened 2026-07-25
fix: stabilize Shadcn Switch and Dialog workflows

guangliang2019:
## Summary

- correct the Shadcn Switch thumb geometry and isolate its shadow from the Root focus ring
- restore Dialog Trigger and Close as transparent semantic wrappers, align the default CloseIcon, and stabilize nested trigger event/focus routing across WC, React, and Vue
- move Shadcn component preset recipes into the prototype library with a generated CLI projection
- add a shared DOM-based Shadcn Dialog conformance journey required for every official Web adapter

## Root cause

The Switch transform distance did not account for the intended symmetric inset, and inherited ring-offset variables leaked into the Thumb shadow.

Dialog nested trigger groups also had three interacting faults: canonical `as-trigger` ancestry was not recognized consistently, semantic root redirection also redirected instance-local `host:focus` and `host:blur`, and synchronous trapped focus movement could run before keyboard modality reached the target. Preset recipes additionally lived only in the CLI registry instead of the prototype library authoring surface.

## Impact

Pointer and keyboard Dialog activation now reach the same behavior owner. Focus-visible remains observable through Cancel → Save changes → CloseIcon cycles, keyboard close restores the nested Trigger Button, and a second interaction cycle no longer loses focus to the page start. Preset structure is authored with the prototype family without introducing a second styling layer.

The shared Web journey is a necessary but insufficient cross-host consistency gate: WC, React, and Vue must all pass the same website demo scenario, while runtime/contract tests and real-browser acceptance continue to cover deeper correctness.

## Validation

- `corepack pnpm@10.32.1 test`
- `corepack pnpm@10.32.1 test:web-conformance`
- `corepack pnpm@10.32.1 check:types:workspace`
- `corepack pnpm@10.32.1 check:agent-doc`
- manual Demo Matrix interaction acceptance for WC, React, and Vue


volkkov (commented):
The happy path is clean. What happens if the third-party API returns 429 here? Might be worth handling that case explicitly.

guangliang2019:
> The happy path is clean. What happens if the third-party API returns 429 here? Might be worth handling that case explicitly.

👀 Thanks! I had a look, but I couldn't find any third-party API call in this PR, so I suspect this comment may have landed here by accident.

Let me know if there's a specific spot you had in mind.

PR #373, opened 2026-08-12
fix(brutalist): enforce preset integrity

### LINKED ISSUE #367: bug(brutalist): overview Button demo loses accent colors and CLI does not guard style-preset mismatch
## Summary

On current `main` (`118d26fd155ca0fcdaa8bb3527cfc6cb6f8e8844`), the Brutalist library overview and the dedicated Button page render the same `demo-brutalist-button` differently:

- Overview: https://www.proto-ui.com/en/ui-libraries/brutalist/#brutalist-button
- Button page: https://www.proto-ui.com/en/ui-libraries/brutalist/components/button/

The dedicated page is correct. On the overview, `mint`, `lavender`, `coral`, and `sky` have transparent backgrounds (and therefore look colorless), while `main` still renders.

This exposes a broader integration risk: a Brutalist component depends on a complete Brutalist style preset, but `proto-ui add` currently only warns when styles are fully disabled. It does not detect an enabled but incompatible preset (for example, the default Shadcn preset followed by `add ... brutalist-button`).

## Browser evidence

Reproduced against current `main` with the Web Components runtime.

| Demo / token | Overview computed background | Button page computed background |
| --- | --- | --- |
| `main` | `rgb(255, 210, 63)` | `rgb(254, 240, 138)` |
| `mint` | transparent | `rgb(167, 243, 208)` |
| `lavender` | transparent | `rgb(221, 214, 254)` |
| `coral` | transparent | `rgb(254, 205, 211)` |
| `sky` | transparent | `rgb(186, 230, 253)` |

On the overview's local demo scope:

- `--pui-main` is defined
- `--pui-mint`, `--pui-lavender`, `--pui-coral`, and `--pui-sky` are empty

The dedicated page defines all of them via `BrutalistPageStyle`.

There is a second visible pair-integrity problem on the overview: `--pui-destructive-foreground` is not locally defined, so the destructive label inherits an unrelated site color rather than the intended black paired foreground.

## Expected behavior

The same demo and runtime should preserve the same Brutalist fill/foreground pairs in every docs surface. This follows the draft catalog contract:

- `P-BRUTALIST-BUTTON-COLOR-PROP`
- `P-BRUTALIST-BUTTON-PAIR-INVARIANT`
- `P-BRUTALIST-BUTTON-LIGHT-DARK`

In particular, a solid color selects a fixed background/foreground pair; missing consumer theme variables must not silently degrade the control to transparent.

For CLI consumers, adding a component whose required `stylePreset` is absent or incompatible should not succeed silently with a visually broken result.

## Root cause

The docs have multiple independently maintained sources for the same theme contract:

1. `apps/www/src/components/BrutalistPageStyle.astro` defines the complete Brutalist palette.
2. `apps/www/src/components/PrototypeLibraryOverview.astro` hand-writes a partial local scope containing `main` but not the other Button accent tokens or all paired foregrounds.
3. The CLI separately generates `brutalist-theme.css`.

The overview is therefore a partial consumer of the token contract. It happens to cover the default `main` example but not the non-default colors.

Separately, the registry correctly declares `stylePreset: 'brutalist'`, but `packages/cli/src/commands/add.ts` only checks `styles.enabled === false`. An enabled `shadcn` preset does not trigger any diagnostic.

### 1. Make the preset manifest canonical and reusable

Define the complete Brutalist semantic-token manifest once (including light/dark values and paired foregrounds), then generate/project from that source into:

- CLI `brutalist-theme.css`
- docs page-level theme CSS
- isolated preview/demo scopes

Avoid another hand-maintained subset in `PrototypeLibraryOverview.astro`. A preview should request `brutalist` as a preset/scope, not enumerate whichever variables its current examples happen to need.

This also keeps future components or new color variants from silently exceeding a preview's incomplete variable list.

### 2. Treat a component's style preset as an install-time dependency

When `proto-ui add` sees `componentEntry.stylePreset`:

- styles disabled: retain a clear actionable diagnostic;
- configured preset matches: proceed;
- configured preset differs or is missing: fail fast (or require an explicit migration/override) with exact remediation.

For the current single-preset config model, the safe minimum is a mismatch error rather than silently changing the whole application's theme. If mixed style families are a supported goal, model them as explicit scoped preset dependencies instead of letting global CSS variables collide.

The CLI should also verify/generate the complete preset token set, not merely record `styles.preset` in config.

### 3. Add behavioral coverage at both boundaries

- Docs/browser test: mount `demo-brutalist-button` in the overview-style isolated scope and assert non-transparent computed backgrounds plus black paired foregrounds for `main|mint|lavender|coral|sky` and destructive.
- Projection parity test: the overview and detail page should resolve the same semantic colors (allowing intentional theme-value differences only if explicitly specified).
- CLI integration test: default/Shadcn init followed by `add <host> brutalist-button` must report the preset mismatch; Brutalist init followed by add must render with all required semantic variables.
- Manifest completeness test: every color utility referenced by the Brutalist prototype catalog resolves to a declared preset variable and paired foreground where required.

## Acceptance criteria

- [ ] Overview Button renders `main`, `mint`, `lavender`, `coral`, `sky`, surface, and destructive with non-transparent intended fills and correct paired foregrounds in light and dark themes.
- [ ] Overview and detail page no longer maintain divergent hand-written Brutalist token subsets.
- [ ] CLI detects enabled-but-incompatible style preset use when adding a Brutalist component.
- [ ] CLI provides an actionable safe remediation without silently replacing an existing application's preset.
- [ ] Automated tests cover computed color resolution, pair integrity, and the CLI mismatch path.

cyjin-yl:
## Summary
- make the Brutalist package own one canonical Light/Dark semantic theme manifest and generate the offline CLI projection from it
- reject enabled style-preset mismatches before package inspection or project mutation, while preserving the explicit `--no-styles` ownership path
- project the canonical theme into all Brutalist documentation pages and remove divergent page-local palette subsets
- add draft decision/test entities and executable CLI, theme, live-theme, generation, and package-surface coverage

## Verification
- `corepack pnpm@10.32.1 install --frozen-lockfile`
- `corepack pnpm@10.32.1 check:package-manifests`
- `corepack pnpm@10.32.1 check:types`
- `corepack pnpm@10.32.1 test` — 281 passed, 3 skipped; 1,252 tests passed, 34 todo
- `corepack pnpm@10.32.1 check:styles:preset`
- `corepack pnpm@10.32.1 check:prototype-catalog`
- `corepack pnpm@10.32.1 check:agent-doc`
- `git diff --check`
- production docs build — 190 pages, 188 Pagefind pages

Closes #367


apps/www/src/components/BrutalistPageStyle.astro
```diff
@@ -1,55 +1,20 @@
...
-    --pui-mint: #a7f3d0;
-    --pui-mint-foreground: #000000;
-    --pui-lavender: #ddd6fe;
-    --pui-lavender-foreground: #000000;
-    --pui-coral: #fecdd3;
-    --pui-coral-foreground: #000000;
-    --pui-sky: #bae6fd;
-    --pui-sky-foreground: #000000;
-  }
+<style is:inline set:html={themeCss}></style>
```
guangliang2019, line 10: This root-level projection is still overridden by the more specific, descendant-level custom properties in `PrototypeLibraryOverview.astro` (`.prototype-library--brutalist .prototype-card__demo`). That block retains the old `main`, `destructive`, and `ring` values and pins Light surface variables even in Dark mode. Please remove that local palette or generate the demo-scope variables from this canonical manifest, then cover the overview's computed Light/Dark values.

guangliang2019 (changes requested):
Thanks for consolidating the Brutalist theme manifest and adding the CLI preset preflight. I found two blockers before this can merge:

1. **DCO is failing.** The current commit (`59d069e`) has no `Signed-off-by:` trailer. Please re-sign/amend the commit so the DCO check passes.
2. **The overview still shadows the canonical theme.** `PrototypeLibraryOverview.astro` continues to assign the old palette directly on `.prototype-library--brutalist .prototype-card__demo`. Those descendant custom-property declarations override the new root-level manifest projection. As a result, `main`, `destructive`, and `ring` still diverge from the manifest, and the demo keeps Light surface values in Dark mode.

Please remove the overview-local palette or project the canonical manifest into that demo scope, and add an overview computed-style regression covering Light/Dark parity and the accent foreground pairs.

The GitHub Actions jobs and Vercel deployment are otherwise passing.

apps/www/src/components/BrutalistPageStyle.astro
cyjin-yl: Fixed in `1bf5d182`: removed the overview-local Brutalist palette so the package manifest remains the sole theme owner, and added a computed-style regression covering every canonical custom property plus the main/destructive/accent foreground pairs in Light and Dark. Fresh verification: 6 focused tests passed; `check:types` reported 135 Astro files with 0 errors/warnings/hints; the production docs build completed; `git diff --check` passed. The amended commit now also carries the required `Signed-off-by` trailer.

guangliang2019 (approved):
The requested changes are addressed in `1bf5d182`: the overview-local palette has been removed, Light/Dark computed-style regression coverage now protects the canonical theme projection, and the amended commit passes DCO. All current GitHub Actions checks and the Vercel deployment are green. Approved.

PR #408, opened 2026-08-15
fix(docs): polish mobile shell follow-up

guangliang2019:
## Summary

- pair the active sidebar background with its matching foreground token so the current page stays legible in light and dark themes
- keep the responsive documentation header on a single 56px row instead of the explicit two-row mobile layout
- move the project-specific adapter selector into the mobile menu at phone widths and synchronize desktop/mobile selector instances without duplicate IDs
- add regression coverage for multi-instance adapter selection

## Context

This is a follow-up to #405, which made the documentation mobile shell usable. The merged layout still left two visible regressions: the current sidebar item could render as light-on-light or dark-on-dark, and the narrow header occupied two navigation rows (96px total).

## Root cause

- Starlight's current-page link color remained in effect while the local override changed only the background to `accent`.
- The mobile header explicitly declared `3.5rem 2.5rem` grid rows and a larger mobile header-height token.
- The adapter selector assumed a single fixed DOM ID, so moving a second instance into the mobile menu required idempotent multi-instance initialization and synchronization.

## Verification

- `corepack pnpm@10.32.1 exec vitest run apps/www/src/components/override/AdapterSelect.test.ts apps/www/src/components/override/LanguageSelect.test.ts apps/www/src/components/override/ThemeProvider.test.ts`
- `corepack pnpm@10.32.1 --filter apps-www check`
- `corepack pnpm@10.32.1 docs:build` (196 pages)
- browser regression matrix at 1440, 1000, 768, 390, and 320px in light and dark themes
- verified 56px header height, no horizontal document overflow, correct mobile menu offset, legible active sidebar state, and no duplicate DOM IDs


cyjin-yl (dismissed):
One browser-visible blocker remains in `apps/www/src/components/override/Header.astro` after collapsing the header to one flex row.

**[P1] Keep the 1024px `lg` breakpoint inside the viewport.** At a 1024×900 viewport on `/en/start-here/what-you-saw/`, the PR renders `document.documentElement.clientWidth === 1024` but `scrollWidth === 1040`; the 165px desktop adapter selector ends at x=1040, so its right edge is clipped and the whole document gains 16px of horizontal overflow. The exact base SHA renders 1024/1024 with no overflow. The regression starts when `lg` reveals the three primary links while the `md` language/adapter controls remain in the same non-wrapping row; it clears at 1040px, which is why the reported 1000px check did not catch it.

Please constrain/shrink the header controls or move the relevant visibility breakpoint so the full single-row surface fits from 1024px onward, and add 1024px to the browser regression matrix.

The selector synchronization itself is sound: all 3 focused tests pass, `apps-www check` reports 139 files with 0 diagnostics, and browser checks at 320/390/768/1000 confirm a 56px header, one visible adapter selector, correct mobile-menu offset, and no overflow. Requesting changes only for the reproducible `lg`-boundary overflow.

cyjin-yl (approved):
Approved on refreshed head `0ca52ffe`. The prior 1024px P1 is fixed. Browser evidence covers 320–1440px, exact 56px header height, no horizontal overflow, mobile menu focus/scroll/selector behavior, no duplicate IDs, and 9.95:1 dark / 16.42:1 light active-item contrast. All refreshed GitHub checks are green.

PR #426, opened 2026-08-16
fix(web-component): project text control focus onto host boundary

### LINKED ISSUE #395: bug(web-component): project native Textarea focus state to the prototype host
## Summary

The Web Component Brutalist Textarea receives real native keyboard focus, but the prototype's focus state and focus-visible styling remain false. React and Vue correctly project the same physical `<textarea>` focus into `data-focus-visible`, exposed state, and the Brutalist ring.

Route: https://www.proto-ui.com/en/ui-libraries/brutalist/components/textarea/

Baseline: `70e2eb1a1bcd9264cf8c08e6ede210f96ee04606`

## Reproduction

1. Select Web Components.
2. Tab from the runtime selector into the textarea.
3. Inspect the inner physical `<textarea>` and the custom-element host/exposes.
4. Repeat in React and Vue.

### Web Components

- `document.activeElement` resolves to the native textarea through the component host
- native textarea `:focus-visible` is `true`
- prototype/host `data-focus-visible` is absent/false
- exposed `focused` and `focusVisible` remain `false`
- computed outline remains transparent
- focus ring box-shadow is absent

Screenshot: `/tmp/omp-sshots-1557549e812c99b4.png`

### React and Vue

- the physical textarea is active and `:focus-visible=true`
- `data-focus-visible=true`
- exposed focus state is true
- computed ring is present

React comparison: `/tmp/omp-sshots-155754c4306c99b6.png`

This is adapter parity failure, not a CSS-token choice: Web Components never projects the native control's focus facts into the state consumed by the style rule.

## Expected behavior

The native HTML `<textarea>` remains the one physical focus target in every runtime. Its focus/blur/focus-visible facts must feed the same prototype state and expose channels, and the Brutalist visual rule must style the physical control without creating a second tab stop.

## Likely ownership boundary

The Web Component adapter renders a custom-element host around an inner native physical control. Focus routing must bridge events/facts from that physical control to the logical prototype instance. The fix belongs in the existing physical-control/focus-target projection path, not in a docs-only `:focus-within` patch and not in an extra `tabindex` on the host.

## Acceptance criteria

- [ ] Tabbing focuses exactly one native textarea in Web Components.
- [ ] Native focus sets prototype `focused` and keyboard focus sets `focusVisible`/`data-focus-visible`.
- [ ] Blur clears those facts without stale state.
- [ ] The Brutalist ring is visibly rendered on the physical textarea in Light and Dark.
- [ ] Pointer focus does not incorrectly claim keyboard focus-visible.
- [ ] Disabled and read-only behavior remains correct and no extra focusable wrapper is introduced.
- [ ] React/Vue behavior remains unchanged.
- [ ] Adapter integration tests cover physical-control focus/blur, exposed state, DOM markers, remount, and runtime teardown.

cyjin-yl:
Closes #395.

## Root cause

Host-bound `host:focus` / `host:blur` listeners are attached to the custom-element boundary, while the WC runtime's natively focusable surface is the physical `<textarea>` / `<input>` (`part="control"`). Native focus/blur do not bubble, so Focus never observed the physical editor; React and Vue use a root focus target and were unaffected.

## Change

- The WC view binds physical `focusin` / `focusout` for its text-control target and routes the trusted native event through `createWebProtoEventRouter().dispatchHostRootEvent(...)`, an adapter-private gated ingress. It does **not** dispatch a second public focus/blur DOM event from the custom-element boundary.
- The private ingress invokes the existing `host:*` listener path and preserves the physical target for Focus. Per-view disposal removes the bridge and router listeners; detach/remount/terminal disposal do not duplicate or leak them.
- `focusSelf` delegates to the physical editor through the existing WC focus override; logical `focused` / `focusVisible` and style markers synchronize in both directions.
- Callback payload types are normalized to declared `focus`/`blur` while retaining the originating physical event as non-enumerable `nativeEvent` context.

## Exclusions / residual risk

- No new Focus Host Capability, Event router change, or cross-adapter generic modality model is introduced.
- The private ingress applies only to the WC runtime's physical text control; React and Vue use the existing root focus target path and are unaffected.
- Screenshot pixels and non-Chromium UA behavior were not independently audited.

## Validation (current head `121684f3`)

- Adapter-base/Web Component/Base Textarea: 21 files / 73 tests passed.
- Brutalist browser journey: 6/6.
- `check:types` — 0 errors/warnings/hints.
- `check:prototype-catalog` — OK.
- `test:public-docs` — passed.
- Live CI/DCO/Vercel checks are green on the exact head.
- Local actionlint was not independently run; the existing CI lanes remain the required evidence.

## Contribution provenance

- **Original rights:** The adapter-private focus ingress, WC integration, tests, and spec/test projection updates are original work created for Proto UI. The DCO signer confirms the right to submit them under the MIT License.
- **Third-party/upstream:** No code or assets were copied or ported. The implementation relies only on public DOM focus/focusin/focusout behavior and existing Proto UI Event/Focus/Text Control contracts.
- **Material AI assistance:** An AI coding assistant (OpenAI Codex/GPT-5.6 family via Oh My Pi) materially assisted with root-cause tracing, implementation, tests, documentation, and review remediation. The human contributor reviewed all diffs and validated physical focus, private event transport, remount/teardown, browser paint, focused tests, types, and catalog checks. No third-party or private code was supplied to the model.
- **Employer/client:** The signer confirms the contribution is not derived from an employer/client proprietary repository and that they have the right to publish it.

### DIFF
--- a/packages/adapters/web-component/src/adapt.ts
+++ b/packages/adapters/web-component/src/adapt.ts
@@ -353,15 +353,16 @@ export function AdaptToWebComponent<TProto extends Prototype<any, any>>(
         this._applier = applier;
         let disposeFocusBridge: (() => void) | null = null;
         if (this._textControlTarget) {
-          // Native focus/blur do not bubble from the physical text control, so
-          // project them onto the custom-element boundary where host-bound
-          // focus listeners are attached.
+          // Native focus/blur do not bubble from the physical text control.
+          // Route the trusted physical event through the adapter-private host
+          // ingress: Proto focus facts update without emitting a second public
+          // native-looking event from the custom-element boundary.
           const control = this._textControlTarget;
-          const onFocusIn = (e: FocusEvent) => {
-            if (e.target === control) thisEl.dispatchEvent(new FocusEvent('focus'));
+          const onFocusIn = (event: FocusEvent) => {
+            if (event.target === control) router.dispatchHostRootEvent('focus', event);
           };
-          const onFocusOut = (e: FocusEvent) => {
-            if (e.target === control) thisEl.dispatchEvent(new FocusEvent('blur'));
+          const onFocusOut = (event: FocusEvent) => {
+            if (event.target === control) router.dispatchHostRootEvent('blur', event);
           };
           thisEl.addEventListener('focusin', onFocusIn);
           thisEl.addEventListener('focusout', onFocusOut);
--- a/packages/adapters/web-component/src/adapt.ts
+++ b/packages/adapters/web-component/src/adapt.ts
@@ -358,11 +358,21 @@ export function AdaptToWebComponent<TProto extends Prototype<any, any>>(
           // ingress: Proto focus facts update without emitting a second public
           // native-looking event from the custom-element boundary.
           const control = this._textControlTarget;
+          const projectFocus = (type: 'focus' | 'blur', nativeEvent: FocusEvent) => {
+            const normalized = new FocusEvent(type, {
+              relatedTarget: nativeEvent.relatedTarget,
+            });
+            Object.defineProperty(normalized, 'nativeEvent', {
+              value: nativeEvent,
+              enumerable: false,
+            });
+            router.dispatchHostRootEvent(type, normalized);
+          };
           const onFocusIn = (event: FocusEvent) => {
-            if (event.target === control) router.dispatchHostRootEvent('focus', event);
+            if (event.target === control) projectFocus('focus', event);
           };
           const onFocusOut = (event: FocusEvent) => {
-            if (event.target === control) router.dispatchHostRootEvent('blur', event);
+            if (event.target === control) projectFocus('blur', event);
           };
           thisEl.addEventListener('focusin', onFocusIn);
           thisEl.addEventListener('focusout', onFocusOut);
--- a/packages/adapters/web-component/src/adapt.ts
+++ b/packages/adapters/web-component/src/adapt.ts
@@ -351,6 +351,25 @@ export function AdaptToWebComponent<TProto extends Prototype<any, any>>(
         currentEventGate = eventGate;
         currentRouter = router;
         this._applier = applier;
+        let disposeFocusBridge: (() => void) | null = null;
+        if (this._textControlTarget) {
+          // Native focus/blur do not bubble from the physical text control, so
+          // project them onto the custom-element boundary where host-bound
+          // focus listeners are attached.
+          const control = this._textControlTarget;
+          const onFocusIn = (e: FocusEvent) => {
+            if (e.target === control) thisEl.dispatchEvent(new FocusEvent('focus'));
+          };
+          const onFocusOut = (e: FocusEvent) => {
+            if (e.target === control) thisEl.dispatchEvent(new FocusEvent('blur'));
+          };
+          thisEl.addEventListener('focusin', onFocusIn);
+          thisEl.addEventListener('focusout', onFocusOut);
+          disposeFocusBridge = () => {
+            thisEl.removeEventListener('focusin', onFocusIn);
+            thisEl.removeEventListener('focusout', onFocusOut);
+          };
+        }
 
         let disposed = false;
         const disposeView = () => {
@@ -360,6 +379,8 @@ export function AdaptToWebComponent<TProto extends Prototype<any, any>>(
           eventGate.dispose();
           unbindLogicalEventTarget(this._instanceToken, router.rootTarget);
           router.dispose();
+          disposeFocusBridge?.();
+          disposeFocusBridge = null;
           applier.clear();
           releaseRenderedChildren();
           if (currentEventGate === eventGate) currentEventGate = null;
--- a/packages/adapters/web-component/src/adapt.ts
+++ b/packages/adapters/web-component/src/adapt.ts
@@ -358,27 +358,22 @@ export function AdaptToWebComponent<TProto extends Prototype<any, any>>(
           // ingress: Proto focus facts update without emitting a second public
           // native-looking event from the custom-element boundary.
           const control = this._textControlTarget;
-          const projectFocus = (type: 'focus' | 'blur', nativeEvent: FocusEvent) => {
-            const normalized = new FocusEvent(type, {
-              relatedTarget: nativeEvent.relatedTarget,
-            });
-            Object.defineProperty(normalized, 'nativeEvent', {
-              value: nativeEvent,
-              enumerable: false,
-            });
-            router.dispatchHostRootEvent(type, normalized);
+          // Bind native focus/blur directly on the known control so the
+          // callback receives the real DOM event object with target/currentTarget
+          // intact. The private transport preserves both the declared type and
+          // the physical event identity without dispatching a second public
+          // boundary event.
+          const onFocus = (event: FocusEvent) => {
+            if (event.target === control) router.dispatchHostRootEvent('focus', event);
           };
-          const onFocusIn = (event: FocusEvent) => {
-            if (event.target === control) projectFocus('focus', event);
+          const onBlur = (event: FocusEvent) => {
+            if (event.target === control) router.dispatchHostRootEvent('blur', event);
           };
-          const onFocusOut = (event: FocusEvent) => {
-            if (event.target === control) projectFocus('blur', event);
-          };
-          thisEl.addEventListener('focusin', onFocusIn);
-          thisEl.addEventListener('focusout', onFocusOut);
+          control.addEventListener('focus', onFocus);
+          control.addEventListener('blur', onBlur);
           disposeFocusBridge = () => {
-            thisEl.removeEventListener('focusin', onFocusIn);
-            thisEl.removeEventListener('focusout', onFocusOut);
+            control.removeEventListener('focus', onFocus);
+            control.removeEventListener('blur', onBlur);
           };
         }
 
--- a/packages/adapters/web-component/test/text-control.integration.test.ts
+++ b/packages/adapters/web-component/test/text-control.integration.test.ts
@@ -59,10 +59,10 @@ const focusTextareaPrototype = definePrototype({
       focusable.focusSelf(options)
     );
     def.event.on('host:focus', (_run, event) => {
-      projectedHostFocusTypes.push(String(event?.type));
+      projectedHostFocusTypes.push(String((event as any)?.type));
     });
     def.event.on('host:blur', (_run, event) => {
-      projectedHostFocusTypes.push(String(event?.type));
+      projectedHostFocusTypes.push(String((event as any)?.type));
     });
     const sync = (props: Readonly<ControlProps>) => {
       control.sync({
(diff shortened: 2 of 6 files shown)


packages/adapters/web-component/test/text-control.integration.test.ts
```diff
@@ -100,4 +130,39 @@ describe('adapter-web-component text control', () => {
...
+    if (!textarea) throw new Error('physical textarea was not materialized');
+    const exposes = shell.getExposes() as {
+      focused: { get(): boolean };
+      focusVisible: { get(): boolean };
+    };
+    expect(exposes.focused.get()).toBe(false);
+    expect(exposes.focusVisible.get()).toBe(false);
+
+    window.dispatchEvent(new KeyboardEvent('keydown', { key: 'Tab' }));
+    textarea.dispatchEvent(new FocusEvent('focusin', { bubbles: true }));
```
guangliang2019: [P2] Please exercise the native focus path here instead of synthesizing `focusin`. This currently verifies only the bridge callback: it never calls `textarea.focus()` / `blur()`, checks `document.activeElement`, asserts `data-focus-visible` on the physical textarea or the Brutalist ring, or covers pointer modality. #395 explicitly requires those behaviors, so a regression in browser focus delivery or surface-marker/style mirroring would still pass. Please add real keyboard- and pointer-focus coverage against Base/Brutalist Textarea and keep the related spec-test mapping traceable.

packages/adapters/web-component/test/text-control.integration.test.ts
```diff
@@ -100,4 +130,39 @@ describe('adapter-web-component text control', () => {
...
+    expect(shell.hasAttribute('data-focus-visible')).toBe(true);
+
+    textarea.dispatchEvent(new FocusEvent('focusout', { bubbles: true }));
+    expect(exposes.focused.get()).toBe(false);
+    expect(exposes.focusVisible.get()).toBe(false);
+    expect(shell.hasAttribute('data-focus-visible')).toBe(false);
+
+    shell.remove();
+    await flush();
+    expect(() =>
```
guangliang2019: [P2] This assertion does not prove teardown is inert. `dispatchEvent()` normally does not throw even if the bridge listener is still attached, and the disposed router/event gate can mask the leak. Count projected `focus` events on `shell` (or spy on `shell.dispatchEvent`) and assert the count does not change after disposal; also cover reconnect/remount so duplicate bridge registration is detectable.

guangliang2019 (dismissed):
The focus bridge implementation direction looks sound, but the regression coverage does not yet prove the end-to-end failure or the teardown acceptance criteria from #395. Please strengthen these tests before merge.

packages/adapters/web-component/test/text-control.integration.test.ts
cyjin-yl: Resolved on `f8042e8e`. The unit regression now drives native `textarea.focus()` / `blur()`, asserts `document.activeElement`, exposed facts, host and physical-control markers, pointer modality, and a single native tab stop. The browser regression tabs from the runtime selector into the physical WC textarea and verifies the computed Brutalist ring in Light and Dark.

packages/adapters/web-component/test/text-control.integration.test.ts
cyjin-yl: Resolved on `f8042e8e`. The regression counts projected boundary `focus` events, proves exactly one additional projection after detach/remount, and proves bubbling `focusin` from each detached physical control cannot increment the count after disposal. The browser and spec-test mappings are included.

packages/adapters/web-component/src/adapt.ts
```diff
@@ -351,6 +351,25 @@ export function AdaptToWebComponent<TProto extends Prototype<any, any>>(
...
         currentRouter = router;
         this._applier = applier;
+        let disposeFocusBridge: (() => void) | null = null;
+        if (this._textControlTarget) {
+          // Native focus/blur do not bubble from the physical text control, so
+          // project them onto the custom-element boundary where host-bound
+          // focus listeners are attached.
+          const control = this._textControlTarget;
+          const onFocusIn = (e: FocusEvent) => {
+            if (e.target === control) thisEl.dispatchEvent(new FocusEvent('focus'));
```
guangliang2019: [P2] Keep this focus projection inside the adapter's private event ingress. `dispatchEvent(new FocusEvent('focus'))` turns one physical focus into a second public, native-looking DOM event; an exact-head Chrome capture observed the trusted `TEXTAREA` focus followed by an untrusted `WC-BRUTALIST-TEXTAREA-ROOT` focus. Document-level capture listeners and analytics can therefore observe or count the same transition twice, and the adapter transport leaks across the current draft `C-TEXT-CONTROL-0001-F` containment boundary. Please route the physical target into the router/event port privately (or bind the internal host-focus route directly to it) and add a negative assertion that Proto focus facts update without a second public boundary `focus` event. The same applies to the `blur` redispatch on line 364.

guangliang2019 (commented):
Refreshed review on `ab00539b`. The two earlier evidence gaps are resolved on the current head: the tests now exercise real native focus and browser paint, and they cover detach/remount teardown without duplicate registration.

One implementation issue remains in the inline comment below. Separately, the contributor-rights gate is still open: the PR body does not include the provenance statement required by [`internal/governance/contribution-provenance.md`](https://github.com/Proto-UI/Proto-UI/blob/main/internal/governance/contribution-provenance.md). Please state whether the contribution is original or derived, confirm the right to submit it, identify sources and licenses if any, disclose material AI assistance and the human validation performed, and say whether third-party or private code was supplied. The green DCO check is necessary, but the policy explicitly treats provenance disclosure as a separate requirement.

Please address the inline event-boundary issue and the provenance disclosure before merge.

packages/adapters/web-component/src/adapt.ts
cyjin-yl: Fixed in `54d3c0b7`. The WC bridge no longer dispatches a synthetic public boundary focus/blur. `createWebProtoEventRouter` exposes an adapter-private `dispatchHostRootEvent` ingress that invokes existing gated host listeners directly with the physical native event. Regression evidence asserts the textarea becomes active and Proto focus facts update while a boundary public focus listener remains exactly zero across keyboard/pointer/remount/teardown. Focused adapter tests 73/73 and browser 6/6 pass.

packages/adapters/web-component/src/adapt.ts
```diff
@@ -351,6 +351,26 @@ export function AdaptToWebComponent<TProto extends Prototype<any, any>>(
...
         this._applier = applier;
+        let disposeFocusBridge: (() => void) | null = null;
+        if (this._textControlTarget) {
+          // Native focus/blur do not bubble from the physical text control.
+          // Route the trusted physical event through the adapter-private host
+          // ingress: Proto focus facts update without emitting a second public
+          // native-looking event from the custom-element boundary.
+          const control = this._textControlTarget;
+          const onFocusIn = (event: FocusEvent) => {
+            if (event.target === control) router.dispatchHostRootEvent('focus', event);
```
guangliang2019, line 367: [P2] Preserve the declared `host:focus` payload type here. This private ingress directly invokes listeners with the original `focusin` event, so an exact-head callback registered for `host:focus` observes `event.type === 'focusin'`; the blur branch similarly exposes `focusout` under `host:blur`. That violates `C-EVENT-0002-I` and `HC-EVENT-BINDING-0001-C`, which require the payload to expose the event type that produced the declared registration dispatch. Focus currently ignores the payload, so the existing tests stay green, but other Event consumers do not have that protection. Please normalize the private payload to `focus`/`blur` (without dispatching another public DOM event), or bind native `focus`/`blur` directly on the known physical control, and add assertions for both callback payload types alongside the zero-public-event check.

guangliang2019 (commented):
Partial refreshed review on `54d3c0b7`. The earlier native-focus/browser and teardown findings remain resolved, the duplicate public boundary focus/blur event is removed, and the PR body now satisfies the provenance disclosure request. Focused adapter/Event tests pass (9/9), as do `check:types` and `check:prototype-catalog`; all current GitHub checks are green.

One new Event-contract mismatch remains in the inline comment. I reproduced it on the exact head with a focused regression: a `host:focus` callback receives `event.type === 'focusin'` (and the blur path analogously carries `focusout`). Please preserve the private transport/no-public-duplication fix while normalizing the callback payload to the declared focus/blur type, and add the corresponding regression.

Repository review policy limits this no-self-assessment run to a partial COMMENT rather than a disposition-changing review. The verified P2 should still be addressed before the existing changes-requested state is cleared.

guangliang2019 (commented):
Partial refreshed review on `121684f3` (canonical input `53e808c6…`, packet key `4fd2fbdc…`).

`PR426-EVENT-TYPE-001` is resolved: the private callback payload now reports `focus` / `blur`, and the no-public-duplication, focus-state, remount, and teardown regressions remain green.

One P2 remains in `packages/adapters/web-component/src/adapt.ts:362`: the replacement `FocusEvent` is never dispatched, so the callback receives `event.target === null`, `event.currentTarget === null`, and a different object from the physical event. The exact-head probe observed the real `TEXTAREA` target only at the non-enumerable `event.nativeEvent.target` on the original `focusin`; the new regression checks only `event.type`, so it misses this host-event object/target loss. That diverges from the existing Web `host:*` contract evidence, where the callback receives the real DOM event object, and can break prototype consumers that use this host-local escape hatch even though Focus itself ignores the payload.

Please preserve both the private transport and declared type while retaining the physical event object/target—binding native `focus` / `blur` directly on the known control is the simplest route—and assert both types, object/target identity, zero public boundary events, remount, and teardown.

Focused validation: Web Component text-control integration plus host-bound router contract, 2 files / 7 tests passed; the exact target probe reproduced `type=focus`, `target/currentTarget=null`, `nativeEvent.type=focusin`, `nativeEvent.target=TEXTAREA`, public boundary focus count 0. Live canonical CI is green. Poppy delivery evidence remains conflicting (successful workflow context versus a failed-deployment bot comment), so I did not use it to clear or expand the code finding.

Repository review eligibility limits this no-current-assessment pass to a partial COMMENT/ABSTAIN packet; it does not clear the existing changes-requested state.

guangliang2019 (approved):
Exact-head approval for f59acd340f7621813d4d65d5d1bc09084246c1d4 (canonical review input 7fceb135…).\n\nAll prior findings are resolved:\n- native-focus/browser-paint and detach/remount/teardown evidence remains present;\n- the physical text-control focus path stays inside the adapter-private Event ingress and emits no duplicate public boundary focus/blur event;\n- PR426-EVENT-TYPE-001 is resolved because host:focus / host:blur callbacks receive the declared focus / blur types;\n- PR426-EVENT-TARGET-002 is resolved by binding native focus / blur directly on the physical control. An isolated exact-head DOM probe confirmed that the callback receives the identical native Event object and observes the physical textarea as both target and in-callback currentTarget;\n- contributor provenance and DCO are complete.\n\nFocused review validation passed: packages/adapters/web-component/test/text-control.integration.test.ts (4/4), including a disposable identity/target/currentTarget assertion. All four review threads are resolved, and all 13 live checks are green on this exact head. The earlier #511 scope concern is gone: #511 has merged and current base 24cfabbe absorbs that commit, leaving this PR focus/Textarea diff coherent.\n\nNo unresolved technical finding or additional maintainer decision remains in the reviewed scope.

PR #442, opened 2026-08-18
fix: dispatch state watchers in callback scope

### LINKED ISSUE #441: [base][dropdown] Keep focus watcher state writes inside callback phase
## Problem

Opening the canonical Base Dropdown Menu demo through the Web Component adapter reports a runtime phase violation:

```text
[Runtime] exec-phase violation: wc-base-dropdown-item op=state(active).set expected=callback actual=unknown protoPhase=updated
```

The menu still opens, so the failure is easy to miss unless browser console errors are part of acceptance.

Discovered while validating #413. This is an existing runtime/prototype behavior defect and must remain separate from that documentation PR.

## Reproduction

1. Build and preview `apps/www`.
2. Open `/en/ui-libraries/base/dropdown-menu/`.
3. Keep the Web Component adapter selected.
4. Activate the “Base actions” trigger with a pointer click.
5. Observe the console error above.

A fresh page has no error before the trigger is activated. The same error is reproducible after activation.

## Current diagnosis

The stack resolves to `packages/prototypes/base/src/dropdown/item.proto.ts`:

- the focus module updates its observed `focused` state;
- `command.focused.watch(...)` receives the focus-loss event;
- the watcher calls `syncActive(...)`;
- `syncActive` calls `active.set(...)` while the item kernel reports `actual=unknown`.

The existing Web Component demo-renderer test opens the menu through keyboard entry and asserts focus/portal behavior, but it does not fail on this browser console error. The failure therefore needs an explicit regression assertion at the adapter/runtime boundary.

## Acceptance criteria

- [ ] Add a focused regression test that fails on the current out-of-phase watcher callback.
- [ ] Make focus-state watcher callbacks that invoke prototype code run in the instance callback scope.
- [ ] Preserve existing Base Dropdown keyboard, pointer, disabled-item, roving-focus, and portal behavior.
- [ ] Verify the canonical Base Dropdown demo opens without console errors in Web Component mode.
- [ ] Run focused focus/runtime/adapter/dropdown tests.
- [ ] Do not change documented Dropdown semantics or mix the fix into #413.

## Scope boundary

If the defect proves to require a new protocol guarantee or a broader watcher-dispatch policy not already implied by the current callback-phase contract, record that contradiction before expanding the implementation scope.

guangliang2019:
## Summary

- add a runtime-owned dispatcher to the privileged State port
- invoke State watch callbacks inside the owning instance's callback scope
- preserve the callback run handle and restore the idle kernel phase afterward
- add a regression contract test for privileged host/module state mutations

## Root cause

Privileged modules can update State slots while the kernel is otherwise idle. State forwarded those mutations to observed/borrowed watchers synchronously, but unlike Context it did not re-enter the owning runtime callback scope. A watcher that performed the contract-allowed update of another state therefore hit an `actual=unknown` phase guard.

This surfaced in the Base Dropdown Menu Web Component demo: opening the menu moves focus from the trigger into an Item, and the Item's `focused.watch(...)` callback synchronizes `active`. The menu still opened, but the browser reported an exec-phase violation.

The fix centralizes the phase boundary in State rather than adding a Dropdown-specific exception. It does not change Dropdown semantics or couple watchers to render/update scheduling.

## Validation

- regression test fails without the State dispatcher and passes with it
- 78 focused State, Focus, Runtime, Dropdown, Web Component, React, and Vue test files passed (264 tests)
- `corepack pnpm@10.32.1 check:types`
- `corepack pnpm@10.32.1 --filter apps-www build`
- browser: Base Dropdown opened, committed, closed, and reopened in Web Component mode with no console errors

Closes #441


packages/modules/state/src/types.ts
```diff
@@ -9,7 +9,16 @@ import type {
...
 
+export type StateCallbackDispatcher = (fn: (ctx: unknown) => void) => void;
+
 export type StatePort = {
+  /**
+   * Runtime-owned bridge that dispatches watch callbacks in the owning
+   * instance's callback scope, including mutations initiated by privileged
+   * module ports while the kernel itself is otherwise idle.
+   */
+  setCallbackDispatcher(dispatch: StateCallbackDispatcher): void;
```
guangliang2019, line 20: [P2] Catalog this new privileged-port and callback-binding guarantee. `M-STATE-0001-B` currently enumerates the State port without this method, while `C-STATE-0010` / `T-STATE-0004` do not project the callback-phase and current-run binding now asserted by the new runtime test. `C-CORE-SYNTAX-0002-E/F` may supply the upstream rule, but the State M/T relations and revision/evidence still need to connect it; otherwise implementation and a contract-named test become a new shadow guarantee. If the existing entities do not authorize this broader policy, record the contradiction as #441's scope boundary requires.

guangliang2019 (commented):
The implementation and focused tests pass, but the new callback-dispatch guarantee needs source-of-truth projection before merge.

cyjin-yl (approved):
完整复核通过，批准。

我重点核对了这次从 Dropdown 症状回收到 State/Runtime 边界的修复是否成立：

1. watcher callback 的 phase 进入由 State port 的统一 dispatcher 完成，`next` 与 terminal `disconnect` 都经过同一路径，没有组件或调用方特判。
2. Runtime 注入的是所属 instance 的 `callbackScope.runNoSync(kernel.run, ...)`；因此 callback 收到当前 run handle，并继承 Context 已使用的异常安全清理、嵌套 depth 与 after-callback flush 语义。
3. contract regression 同时覆盖了 idle `unknown` → `callback`、callback 内合法 state mutation、run identity、正常返回后的 phase 恢复，以及 watcher 抛错后的 phase/context 恢复；这正好约束了原根因，而不是只验证 Dropdown 不报错。
4. `C-STATE-0010-F/G`、`M-STATE-0001-B` 与 `T-STATE-0004-CASE-CALLBACK-SCOPE` 已把实现、模块职责和可执行证据闭环，也明确没有把 watch 变成隐式 render/update 系统。
5. 当前 head 的 GitHub Actions CI 已成功。

一个非阻塞的后续观察：若未来 State 支持批量 watcher transaction，需要重新决定“每个 watcher 各自进入 callback scope并在最外层结束时 flush”与“整批只进入一次”的调度粒度；本 PR 沿用 Context 的现有 dispatcher 模式，对当前同步 watcher contract 是一致的。

packages/modules/state/src/types.ts
cyjin-yl: 已在当前 head `cac8cccd` 完成这条 source-of-truth 投射：`C-STATE-0010` 增加 callback-scope/current-run 与 phase-restoration 准则，`M-STATE-0001-B` 纳入 privileged dispatcher，`T-STATE-0004` 增加并映射 runtime contract case，同时补齐 revisions 与 `C-CORE-SYNTAX-0002-E/F` 依赖。独立复核通过，关闭此线程。

PR #436, opened 2026-08-16
feat(shadcn): project Base Textarea

HyacinthHaru:
## Summary

Projects the host-mediated Base Textarea into `@proto.ui/prototypes-shadcn`. Base keeps the single host-owned editor and the whole editing lifecycle; the projection adds the field surface, focus ring, and disabled treatment only.

## Related context

- Issue: closes #382
- Applicable spec entities and lifecycle: `P-SHADCN-TEXTAREA` and `T-SHADCN-TEXTAREA-0001`, both new and `draft` at `0.3.0-alpha.0`. `P-BASE-TEXTAREA` is unchanged.
- Criteria / test anchors: every `P-SHADCN-TEXTAREA-*` criterion maps to a `T-SHADCN-TEXTAREA-0001-CASE-*` case; `sources` pins the upstream file.
- Related upstream: shadcn-ui/ui revision `f31ed81983653919dd4fe77aee4b4859f610f1dc`, `apps/v4/registry/new-york-v4/ui/textarea.tsx` — the same pinned baseline as the Switch and Separator entities.

## Scope boundary

**Already decided by the issue.** Consume Base Textarea and materialize exactly one host-owned editor; own classes, sizing, border, focus ring, disabled, and invalid presentation *only where those states already exist*; keep the renderer contentless; deliver the source, `P-SHADCN-TEXTAREA`, exact exports, CLI registry, focused tests, EN/ZH docs, a real demo, preview registry, and preset/token closure. No Base Input, auto-grow, rich text, form submission, or validation orchestration.

**Decided here — three upstream omissions, each on existing precedent.**

Base Textarea exposes exactly six states: `value`, `disabled`, `readOnly`, `focused`, `focusVisible`, `composing`. There is no `invalid`. So the issue's "invalid presentation only where those states already exist" resolves to *no invalid presentation*.

| upstream | decision | why |
| --- | --- | --- |
| `aria-invalid:ring-destructive/20`, `aria-invalid:border-destructive` | omitted | The CSS service does support `aria-*` variants, so this is expressible — but with no Base `invalid` state it would be presentation with no protocol behind it, driven solely by a host-set ARIA attribute. `P-SHADCN-BUTTON` already records `aria-invalid` selectors as a compatibility difference rather than implementing them; this follows that precedent. |
| `field-sizing-content` | omitted | That token *is* auto-resize, which the issue's Out of scope section excludes. |
[...]

PR #447, opened 2026-08-18
test(cli): cover commas in joined arbitrary tokens

HyacinthHaru:
## Summary

Adds the CLI scanner unit case I said I would open separately once #436 landed: an array literal joined into a class string must keep a comma-containing arbitrary token whole.

Test-only. One file, one new `it()`. No source, spec, or generated output changes.

## Related context

- Follow-up to #436, where the scanner fix shipped without its own focused case. The fix lives in `packages/cli/src/services/prototype-style-tokens.ts` — array literals kept an `elements` list so `.join()` re-joins elements instead of re-splitting a comma-joined string.
- `packages/cli/test/prototype-style-tokens.test.ts` is not registered as a `T` implementation, so no spec entity changes.

## Why this case and not a broader one

The scanner already had cases for template interpolation, cross-file constants, and the four rule-lowering shapes. The uncovered edge was narrow and specific: `transition-[color,box-shadow]` is the only token in the tree where a comma is *inside* a token rather than between two, and it only reaches the array path through the Textarea's `[...].join(' ')` shape. The case reproduces exactly that shape.

## Verification

- Reverted both fix lines locally and re-ran: the new case fails with `expected [ 'box-shadow]', 'duration-150', …(2) ] to include 'transition-[color,box-shadow]'`; the other five stay green. Restored, all six green.
- `vitest run packages/cli/test` — 40 passed.
- `check:prototype-catalog`, `check:styles:preset`, `check:release-version`, `check:types:workspace` — OK.
- `prettier --check` on the touched file — clean.

## Source-of-truth alignment

### DIFF
--- a/packages/cli/test/prototype-style-tokens.test.ts
+++ b/packages/cli/test/prototype-style-tokens.test.ts
@@ -186,4 +186,36 @@ describe('collectProtoStyleTokens', () => {
     expect(tokens).toContain('data-[busy]:opacity-50');
     expect(tokens).toContain('data-[busy]:cursor-wait');
   });
+  it('keeps commas inside arbitrary tokens when an array literal is joined', async () => {
+    await writeFile(
+      path.join(dir, 'field.proto.ts'),
+      [
+        "import { definePrototype, tw } from '@proto.ui/core';",
+        '',
+        'const BASE_TOKENS = [',
+        "  'flex',",
+        "  'transition-[color,box-shadow]',",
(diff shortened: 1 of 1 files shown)


guangliang2019 (approved):
Reviewed the focused regression case against #436. The test reproduces the joined-array path, fails when the scanner fix is reverted, and passes with the current implementation. Focused tests, workspace types, and CI are green.

PR #450, opened 2026-08-18
fix(style): order lowered variants like the generated stylesheet

### LINKED ISSUE #448: bug(style): multi-condition rules paint nothing when runtime and CLI disagree on variant order
## Summary

A rule whose `when` combines three or more conditions can render nothing at all. The runtime joins the lowered variants in the order the conditions were authored, while the CLI extractor sorts them before writing the stylesheet. When those two orders differ, the class the runtime puts on the element has no matching CSS rule, because `[data-pui-style~="…"]` matches whole words.

On current `main` this silently removes the entire hover treatment from the Brutalist Tabs trigger.

Baseline: `a30b02ce`

Route: https://www.proto-ui.com/en/ui-libraries/brutalist/components/tabs/

## Browser evidence

Hovering an unselected trigger sets the state correctly but paints nothing:

```
data-hovered   true
data-selected  false
background-color   rgba(0, 0, 0, 0)
box-shadow         none
transform          none
border-top-color   rgba(0, 0, 0, 0)
```

The resting and hovered screenshots of the tab list are byte-identical (same 5299-byte PNG, deviceScaleFactor 2):

| resting | hovered |
| --- | --- |
| resting | hovered |

## Root cause

The five tokens the runtime puts on the element are absent from every stylesheet on the page, while the same five exist in the generated CSS under a different segment order:

| on the element (runtime) | in the generated CSS and manifest |
| --- | --- |
| `data-[hovered]:not-[data-selected]:not-[data-pressed]:bg-background` | `data-[hovered]:not-[data-pressed]:not-[data-selected]:bg-background` |
| `…:border-black` | `…:border-black` |
| `…:-translate-x-px` | `…:-translate-x-px` |
| `…:-translate-y-px` | `…:-translate-y-px` |
| `…:shadow-[4px_4px_0_0_#000]` | `…:shadow-[4px_4px_0_0_#000]` |

- `packages/modules/rule-expose-state-web/src/create.ts:254` — `variants.join(':')`, author order, no sort.
- `packages/cli/src/services/prototype-style-tokens.ts` — `return variants.sort(compareVariants)`, where `compareVariants` ranks `dark, hover, active, focus, focus-visible, disabled` first and falls back to `localeCompare` for everything else, so `data-[hovered]`, `not-[data-active]`, `not-[data-pressed]`, `not-[data-selected]` sort alphabetically.

Two-condition rules on the same page are unaffected because sorting a pair often preserves the authored order: `data-[selected]:not-[data-pressed]:shadow-[3px_3px_0_0_#000]` is present in the CSS and does paint.

## Scope

Confirmed dead: the five Tabs trigger tokens above.

Confirmed still working: the Brutalist Toggle three-condition family (`data-[hovered]:not-[data-active]:not-[data-pressed]:*`) paints on hover (`box-shadow: rgb(0,0,0) 4px 4px 0 0`, `transform: matrix(1,0,0,1,-1,-1)`, `border-top-color: rgb(0,0,0)`), because its authored order already matches the sort.

So the defect is not "three conditions never work"; it is "the two sides disagree about order, and nothing detects the disagreement". Any rule is one reordering of its `when` away from silently losing its presentation.

## Expected behavior

The class name the runtime writes onto the host is the same string the extractor writes into the stylesheet, for any number of conditions and any authoring order.

## Why tests did not catch it

`packages/prototypes/brutalist/test/tabs.test.ts:139-147` asserts these five tokens as strings on `data-pui-style`. That assertion passes exactly as well when no CSS rule exists, so `T-BRUTALIST-TABS-0001-CASE-3` reports `P-BRUTALIST-TABS-TRIGGER-INTERACTION` as covered while the interaction does not happen.

## Acceptance criteria

- [ ] Variant order is produced by one shared ordering, consumed by both the runtime lowering and the CLI extractor.
- [ ] Hovering an unselected Brutalist Tabs trigger paints the authored fill, border, offset and shadow in Web Components, React and Vue.
- [ ] A test fails when the two sides disagree, rather than asserting the token string alone.
- [ ] Existing multi-condition tokens that currently paint keep painting, with the Toggle family as the regression anchor.

HyacinthHaru:
## Summary

Closes #448.

The runtime joined a rule's lowered variants in the order the conditions were authored; the CLI extractor sorted them before writing the stylesheet. `data-pui-style~="…"` matches whole words, so whenever the two orders differed the class on the host had no rule at all. Both sides now use one ordering, and a test pins them against each other.

## What changed

- `packages/modules/rule-expose-state-web/src/create.ts` — sorts variants before joining, and exports `compareLoweredVariants`.
- `packages/cli/src/services/prototype-style-tokens.ts` — exports its existing `compareVariants` so the two can be compared; behaviour unchanged.
- `packages/cli/test/lowered-variant-order.test.ts` — new. Asserts the two orderings agree, that authoring order does not change the result, and pins the Tabs trigger class the stylesheet actually carries.
- `packages/cli/package.json` — the module added as a devDependency for that test. The published dependency surface is untouched.

The CLI keeps its own copy rather than importing the module, because it ships with only `prompts` and `typescript` and that seemed worth preserving. The new test is what keeps the copies honest.

## Browser evidence

Hovering an unselected Brutalist Tabs trigger, same page, before and after:

| | before | after |
| --- | --- | --- |
| `background-color` | `rgba(0, 0, 0, 0)` | `rgb(245, 245, 245)` |
| `box-shadow` | `none` | `rgb(0, 0, 0) 4px 4px 0 0` |
| `transform` | `none` | `matrix(1, 0, 0, 1, -1, -1)` |
| `border-top-color` | `rgba(0, 0, 0, 0)` | `rgb(0, 0, 0)` |
| screenshots | byte-identical | differ |

Before, resting and hovered:

| resting | hovered |
| --- | --- |
| resting | hovered |

## Two test changes worth reviewing closely

**`packages/prototypes/brutalist/test/tabs.test.ts`** asserted the five authoring-order token strings. Those strings are the ones with no CSS rule, so the case reported `P-BRUTALIST-TABS-TRIGGER-INTERACTION` as covered while the interaction did not happen. Now it asserts the strings the stylesheet carries.

**`rule.expose-state-web.optimize.v0.contract.test.ts`** pinned `data-[btn-disabled]:dark:bg-zinc-950`. I changed it to `dark:data-[btn-disabled]:bg-zinc-950`, which is what the extractor emits and therefore the only one with a rule. I checked before touching a v0 contract test: neither `C-FEEDBACK-STYLE-0004` nor `C-RULE-EXTENSION-0001` has a criterion about variant order — the first governs author-channel purity, the second equivalent execution — so the old string reads as an incidental pin of the implementation rather than a normative guarantee. If you consider the order normative in the other direction, the fix belongs in the extractor instead and I will redo it that way.

## Scope check

I looked for other multi-condition rules before assuming this was general. The Brutalist Toggle family `data-[hovered]:not-[data-active]:not-[data-pressed]:*` already painted, because its authoring order happened to match the sort — verified on the page (`box-shadow: rgb(0,0,0) 4px 4px 0 0`, `transform: matrix(1,0,0,1,-1,-1)`). So the confirmed breakage was the five Tabs tokens; the change makes the agreement structural rather than accidental.

## Checks

- `vitest run packages/modules packages/prototypes packages/cli` — 555 passed.
- `vitest run packages/adapters` — 467 passed.
- `check:prototype-catalog`, `check:styles:preset`, `check:release-version`, `check:package-manifests`, `check:types:workspace` — all OK.

## Source-of-truth alignment

### DIFF
--- a/packages/cli/src/services/prototype-style-tokens.ts
+++ b/packages/cli/src/services/prototype-style-tokens.ts
@@ -650,7 +650,10 @@ function collectTwTokens(node, scope) {
   }
 }
 
-function compareVariants(a, b) {
+// Twin of `compareLoweredVariants` in
+// `packages/modules/rule-expose-state-web/src/create.ts`. The two are pinned
+// against each other by `packages/cli/test/lowered-variant-order.test.ts`.
+export function compareVariants(a, b) {
   const order = ['dark', 'hover', 'active', 'focus', 'focus-visible', 'disabled'];
   const ai = order.indexOf(a);
   const bi = order.indexOf(b);
--- a/packages/cli/src/services/prototype-style-tokens.ts
+++ b/packages/cli/src/services/prototype-style-tokens.ts
@@ -5,6 +5,8 @@ import fs from 'node:fs/promises';
 import path from 'node:path';
 import ts from 'typescript';
 
+import { canonicalizeLoweredVariants } from '../generated/lowered-variant-order';
+
 export async function collectProtoStyleTokens(root) {
   const files = await collectSourceFiles(root);
   const tokens = new Set();
@@ -517,9 +519,9 @@ function analyzeWhenVariants(node, scope) {
   const out = new Set();
 
   visit(node);
-  const variants = Array.from(out);
+  const variants = canonicalizeLoweredVariants(Array.from(out));
   if (variants.length > 0 && variants.every(isNegativeDataVariant)) return [];
-  return variants.sort(compareVariants);
+  return variants;
 
   function visit(current) {
     if (ts.isArrowFunction(current) || ts.isFunctionExpression(current)) {
@@ -648,17 +650,6 @@ function collectTwTokens(node, scope) {
 
     ts.forEachChild(current, (child) => visit(child, currentScope));
   }
-}
-
-// Twin of `compareLoweredVariants` in
-// `packages/modules/rule-expose-state-web/src/create.ts`. The two are pinned
-// against each other by `packages/cli/test/lowered-variant-order.test.ts`.
-export function compareVariants(a, b) {
-  const order = ['dark', 'hover', 'active', 'focus', 'focus-visible', 'disabled'];
-  const ai = order.indexOf(a);
-  const bi = order.indexOf(b);
-  if (ai !== -1 || bi !== -1) return (ai === -1 ? 999 : ai) - (bi === -1 ? 999 : bi);
-  return a.localeCompare(b);
 }
 
 function isPropertyNamed(node, name) {
--- a/packages/modules/rule-expose-state-web/src/create.ts
+++ b/packages/modules/rule-expose-state-web/src/create.ts
@@ -136,6 +136,22 @@ function isNegativeDataVariant(variant: string): boolean {
   return /^not-\[data-[a-zA-Z0-9-]+\]$/.test(variant);
 }
 
+const VARIANT_ORDER = ['dark', 'hover', 'active', 'focus', 'focus-visible', 'disabled'];
+
+/**
+ * The stylesheet is generated ahead of time by the CLI extractor, which sorts a
+ * rule's variants before writing the class name. `data-pui-style~="…"` matches
+ * whole words, so a class joined in authoring order simply has no rule to hit.
+ * Both sides have to agree, and this is the side that can be changed at runtime.
+ * `packages/cli/src/services/prototype-style-tokens.ts` holds the twin.
+ */
+export function compareLoweredVariants(a: string, b: string): number {
+  const ai = VARIANT_ORDER.indexOf(a);
+  const bi = VARIANT_ORDER.indexOf(b);
+  if (ai !== -1 || bi !== -1) return (ai === -1 ? 999 : ai) - (bi === -1 ? 999 : bi);
+  return a.localeCompare(b);
+}
+
 class RuleExposeStateWebImpl extends ModuleBase {
   private readonly rulePort: RulePort<any>;
   private readonly exposeStateWeb: ExposeStateWebPort;
@@ -251,7 +267,7 @@ class RuleExposeStateWebImpl extends ModuleBase {
       if (!ok || variants.length === 0) continue;
       if (variants.every(isNegativeDataVariant)) continue;
 
-      const prefix = variants.join(':');
+      const prefix = [...variants].sort(compareLoweredVariants).join(':');
       const tokens = c.tokens.map((t) => `${prefix}:${t}`);
       const handle: StyleHandle = { kind: 'tw', tokens };
       this.feedbackPort.useStyleUnsafe(handle);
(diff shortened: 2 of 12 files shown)


packages/modules/rule-expose-state-web/src/create.ts
```diff
@@ -251,7 +267,7 @@ class RuleExposeStateWebImpl extends ModuleBase {
       if (!ok || variants.length === 0) continue;
       if (variants.every(isNegativeDataVariant)) continue;
 
-      const prefix = variants.join(':');
+      const prefix = [...variants].sort(compareLoweredVariants).join(':');
```
guangliang2019: [P2] Match the extractor's deduplication before joining. `analyzeWhenVariants()` accumulates into a `Set`, but this runtime path keeps every lowered condition in an array. Repeating a condition—or two state handles lowering to the same variant—therefore makes the CLI emit one segment while the host gets duplicates such as `data-[hovered]:data-[hovered]:…`, which again has no generated rule. Please canonicalize the full variant list (dedupe + order) on both sides and add a duplicate-lowering regression case.

packages/cli/src/services/prototype-style-tokens.ts
```diff
@@ -650,7 +650,10 @@ function collectTwTokens(node, scope) {
   }
 }
 
-function compareVariants(a, b) {
+// Twin of `compareLoweredVariants` in
+// `packages/modules/rule-expose-state-web/src/create.ts`. The two are pinned
+// against each other by `packages/cli/test/lowered-variant-order.test.ts`.
+export function compareVariants(a, b) {
```
guangliang2019: [P2] This remains a second implementation, and the finite `VARIANT_LISTS` sample cannot detect a later ordering change or new ranked variant that is absent from the fixture. That leaves the original drift class intact and does not meet #448's shared-ordering acceptance criterion. It also requires exporting `compareLoweredVariants` through the public `@proto.ui/module-rule-expose-state-web` barrel solely for the test. Please derive both consumers from one canonical internal source (or a generated artifact with exhaustive parity) without widening the published module API.

guangliang2019 (changes requested):
The current Tabs regression is fixed and the focused tests pass, but the extractor/runtime identity is still not structural. Please close the two remaining mismatch paths below before merging.

guangliang2019 (approved):
复查 `af46f55f`：上一轮两个阻塞点均已关闭。

- runtime 与 CLI 现在都通过同一份 canonical source 生成的 `canonicalizeLoweredVariants` 完成去重和排序；重复 lowering 的真实 contract regression 已覆盖。
- variant ordering 已收敛为单一生成源，并由 `check:styles:variant-order` 做字节级一致性检查；内部 comparator 不再通过模块 barrel 暴露。

本地独立复测通过：`check:styles:variant-order`、3 个相关测试文件（19 tests）以及 `check:types:workspace`；当前 GitHub checks 也全部通过。

非阻塞：PR 描述仍保留上一版“双实现/公开导出”的说明，合并前可顺手同步为当前生成式设计。

PR #457, opened 2026-08-20
test: share one dev server across the browser suites

HyacinthHaru:
## Summary

Per your call on #453: option 1, as its own PR.

`test:runtime` now starts one `apps-www` dev server, waits for it, and hands its URL to `vitest run` through `PROTO_UI_BROWSER_BASE_URL`. The three browser suites keep the code that spawns their own server, so running a file on its own still works.

## Why it raced

The suites each spawn a server, and all three point at the same `apps/www` root, so they share `apps/www/.astro`. `test:runtime` is `vitest run`, which runs files in parallel, so two Astro content stores write `data-store.json.tmp` and race the rename — the loser gets `ENOENT` and its dev server exits.

## Measured, before and after

Sampling `pgrep -f 'astro dev'` every 10s across a full run of the three suites:

| | before | after |
| --- | --- | --- |
| peak concurrent dev servers | 2 | **1** |
| leftover after the run | 1 | **0** |
| wall clock | 33s | **21s** |
| result | 1 suite failed | **5 files, 13 tests passed** |

The 12s comes from not starting two extra servers.

## One thing that was not obvious

My first attempt still showed two servers. The three suites wait on **different** readiness routes — base textarea, brutalist switch and tabs, shadcn textarea — and Astro dev compiles a route on first request. A suite probing a cold route blew its own 10s hook timeout, decided the external server was unusable, and spawned its own. The script now warms all four routes before handing the URL over. Worth knowing if a suite ever adds a route.

## Cleanup paths

- **Normal and failing runs** — `finally` stops the server; vitest's exit code is passed through.
- **Signals** — `SIGINT`/`SIGTERM`/`SIGHUP` stop the server and exit 130/143. Verified: `kill -TERM` on the orchestrator mid-run left `pgrep -f 'astro dev'` at 0 and exited 143.
- The child is detached into its own process group and signalled as a group, so the dev server's own children go with it. `SIGTERM` first, `SIGKILL` after 5s.

## Verification

- `node scripts/test/run-runtime-tests.mjs apps/www/src/content/docs/zh-cn` — 5 files, 13 tests, exit 0, peak 1 server, 0 leftover.
- `pnpm vitest run apps/www/src/content/docs/zh-cn/demo-shadcn-controls.browser.test.ts` with no env var — 3 tests pass through the fallback, 0 leftover. The fallback is intact.
- No browser assertion changed; no test file is touched by this PR.

## Source-of-truth alignment

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "test:release": "node --test scripts/release/test/*.test.mjs",
     "test:types": "echo \"[types] checking type contracts...\" && tsc -p internal/contracts/types/tsconfig.json && echo \"[types] OK\"",
     "test:web-conformance": "vitest run packages/web-conformance/test",
-    "test:runtime": "vitest run",
+    "test:runtime": "node scripts/test/run-runtime-tests.mjs",
     "styles:variant-order:generate": "node --import tsx scripts/styles/generate-lowered-variant-order.ts",
     "check:styles:variant-order": "node --import tsx scripts/styles/generate-lowered-variant-order.ts --check"
   },
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "test:release": "node --test scripts/release/test/*.test.mjs",
     "test:types": "echo \"[types] checking type contracts...\" && tsc -p internal/contracts/types/tsconfig.json && echo \"[types] OK\"",
     "test:web-conformance": "vitest run packages/web-conformance/test",
-    "test:runtime": "node scripts/test/run-runtime-tests.mjs",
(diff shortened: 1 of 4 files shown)


cyjin-yl (approved):
Maintainer takeover final review: approved at `ab8e8d24`. The takeover corrected the focused-argument forwarding bug and narrowed the shared Astro lifetime to the only three suites that consume it. Local exact-head evidence: plan contract 2/2; focused fixture exactly 1 file / 2 tests; full runtime split 301 passed + 3 skipped (1,323 passed, 34 todo) followed by 3 browser files / 9 tests, zero failures, 133 s. GitHub's refreshed head is fully green including DCO, test, type-check, package build, release stage/consumer, CLI smoke, and Vercel.

Issue #455, opened 2026-08-19
bug(style): focus ring repaints on every descendant that carries a shadow token


HyacinthHaru:
## Summary

A focus ring paints on every descendant that carries any shadow token, not only on the element that has focus. The ring variables are custom properties, so they inherit, and each shadow token re-composes them.

The Brutalist Switch shows it plainly: with real keyboard focus the 20px thumb grows its own copy of the root's ring.

Baseline: `d7852bad`

Route: https://www.proto-ui.com/en/ui-libraries/brutalist/components/switch/

## Browser evidence

Tab into the Switch from the previewer's runtime select. The root reports `data-focus-visible`, and root and thumb come back **byte-identical**, ring layers included:

```
root : rgb(245,245,245) 0 0 0 2px, rgb(23,23,23) 0 0 0 4px, rgb(0,0,0) 3px 3px 0 0
thumb: rgb(245,245,245) 0 0 0 2px, rgb(23,23,23) 0 0 0 4px, rgb(0,0,0) 3px 3px 0 0
```

The thumb never declares a ring. Its only shadow token is `shadow-[3px_3px_0_0_#000]`. Reproduced in Light and Dark, in Web Components, React and Vue.

| light | dark |
| --- | --- |
| ![light](https://raw.githubusercontent.com/HyacinthHaru/Proto-UI/audit-artifacts/2026-08-18-visual-audit/switch-ring-leak-light.png) | ![dark](https://raw.githubusercontent.com/HyacinthHaru/Proto-UI/audit-artifacts/2026-08-18-visual-audit/switch-ring-leak-dark.png) |

## Root cause

`packages/cli/src/services/proto-style-css.ts:647` — every shadow token emits:

```css
box-shadow: var(--pui-ring-offset-shadow, 0 0 #0000),
            var(--pui-ring-shadow, 0 0 #0000),
            var(--pui-shadow, 0 0 #0000);
```

and `:639-640` sets those two variables on whatever element wears `ring-*`.

The `0 0 #0000` fallback only applies when the property is **unset**. Once an ancestor sets it, it inherits, the fallback never fires, and any descendant that composes a box-shadow paints the ancestor's ring again at its own size.

Upstream Tailwind avoids this with a universal reset — `*, ::before, ::after { --tw-ring-shadow: 0 0 #0000; … }`. This generator has no equivalent: there is no universal-selector rule in the emitted stylesheet at all, and no `--pui-ring-shadow` reset anywhere in the generator.

[...]

PR #463, opened 2026-08-21
fix(style): use the theme background for implicit ring offsets

### LINKED ISSUE #456: bug(style): ring offset falls back to white and paints a band in Dark
## Summary

The ring offset falls back to a hardcoded `#fff`. In Dark mode that paints a solid white band around the focused control — the opposite of a gap, and the brightest thing on the page.

Baseline: `d7852bad`

## Browser evidence

shadcn Dropdown Menu trigger, Dark, reached by keyboard so `data-focus-visible` is set:

```
box-shadow: rgb(255, 255, 255) 0 0 0 2px,
            oklab(0.556 … / 0.5) 0 0 0 5px,
            rgba(0, 0, 0, 0.05) 0 1px 2px 0
--pui-ring-offset-color: (unset, so #fff)
tokens: data-[focus-visible]:ring-3 data-[focus-visible]:ring-ring/50 data-[focus-visible]:ring-offset-2
```

white ring offset in dark

## Root cause

`packages/cli/src/services/proto-style-css.ts:639`:

```css
--pui-ring-offset-shadow: 0 0 0 var(--pui-ring-offset-width, 0px) var(--pui-ring-offset-color, #fff);
```

`--pui-ring-offset-color` is only set by `ring-offset-background` (`:201`) or an explicit ring-offset colour (`:489`). A prototype that writes `ring-offset-2` alone gets `#fff` — invisible against Light paper, a white box against every Dark surface.

## Scope

Eleven prototype sites write `ring-offset-*`. Exactly three omit `ring-offset-background`, and all three were found independently while auditing different components:

```
packages/prototypes/brutalist/src/hover-card/trigger.proto.ts:47
packages/prototypes/shadcn/src/dropdown/trigger.proto.ts:113
packages/prototypes/shadcn/src/hover-card/trigger.proto.ts:34
```

The other eight, including `BRUTALIST_FOCUS_TOKENS` in `brutalist/src/style.ts:2`, do write it and render correctly.

## Two ways to fix, and I would rather you pick

**Add the token at the three sites.** Smallest diff, leaves the trap in place — the twelfth site to write `ring-offset-2` will land on white again, and nothing fails.

**Make the default theme-relative** — `var(--pui-ring-offset-color, var(--pui-background))`. Removes the whole class, and the eight sites that already write `ring-offset-background` resolve to the same value, so their rendering is unchanged. It does change a shared default that every consumer inherits, which is why I am asking rather than assuming.

I lean toward the second with the three sites left as-is afterwards, but this is a shared-default change and the call is yours.

## Expected behavior

A focused control's ring offset reads as a gap against the surface behind it in both themes, whether or not the prototype spells the offset colour out.

## Acceptance criteria

- [ ] The three sites above render a theme-appropriate offset in Dark, in Web Components, React and Vue.
- [ ] The eight sites that already declare `ring-offset-background` render unchanged.
- [ ] A test fails when a ring offset resolves to a colour unrelated to the surface behind it, rather than asserting token strings.

## Context

Found by the audit that produced #448, #449, #454 and #455. Related to #455 in file and mechanism — both are about how `ringShadow()` and `composedShadow()` compose — but they are separate defects and can be fixed independently.

cyjin-yl:
## Summary

Fixes #456.

- Default an unset ring-offset colour to `var(--pui-background)` instead of the hardcoded `#fff`.
- Leave the three affected prototype sites unchanged so the shared default removes the trap for future `ring-offset-*` consumers as well.
- Preserve the existing explicit `ring-offset-background` utility.
- Add a real Chromium regression that keyboard-focuses the shadcn Dropdown Menu trigger in Dark mode and compares the painted offset layer with the control surface in Web Components, React, and Vue.

## Related context

- Issue: #456
- Applicable spec entities and lifecycle: `C-FEEDBACK-STYLE-0003` and `P-SHADCN-DROPDOWN-MENU-TRIGGER` are draft; this restores their existing style-intent/focus-feedback direction rather than widening the public protocol.
- Criteria / test anchors:
  - `packages/cli/test/ring-offset-default.test.ts`
  - `apps/www/src/content/docs/zh-cn/demo-ring-offset-default.browser.test.ts`
- Related upstream: none; this is a Proto UI physical CSS translation default.

## Scope boundary

- The issue had already narrowed the choice to adding three local tokens or making the shared default theme-relative, and preferred the shared-default direction.
- This pull request chooses the theme-relative default and adds generated-CSS plus painted-browser evidence.
- Explicit offset colours, ring widths, ring colours, shadows, prototype token lists, and theme values are unchanged.
- The inherited-ring recomposition defect in #455 is intentionally out of scope.

## Source-of-truth alignment

- [ ] Normative entity revisions are not applicable: this fixes physical translation so existing focus-feedback intent is rendered correctly; it does not add or widen a P/C/T guarantee.

## Contribution provenance

Select every applicable item. Original, third-party, AI-assisted, and employer or client considerations may coexist.

- [x] This contribution was created for this repository and the submitting account has the right to submit it.
- [ ] This contribution adapts or derives from third-party material.
- [ ] This contribution may be subject to employer or client ownership, and I have confirmed that I may submit it.

### Third-party or upstream sources

| Source | Version / commit | License | Material used | Required attribution |
| ------ | ---------------- | ------- | ------------- | -------------------- |
| None | — | — | — | — |

### AI assistance

OpenAI's connected coding agent assisted with issue triage, source tracing, implementation, and test authoring. The resulting diff was reviewed against #456, the applicable draft spec entities, the three affected prototype sources, the existing browser harness, and a `main...head` branch comparison. No private third-party source material was provided.

## DCO

Every commit in this pull request contains a `Signed-off-by: cyjin-yl <cyjin.yl@gmail.com>` trailer.

- [x] I have checked the DCO status of this PR.

## Prototype website preview

- [ ] This pull request adds or changes public Prototype behavior and includes the required reachable website page or page update.
- [x] A new or updated website page is not applicable because the shared physical-CSS fix is exercised by existing public Dropdown Menu and Hover Card pages; the regression uses the existing real shadcn Dropdown Menu page.

- Public docs route: `/en/ui-libraries/shadcn/dropdown-menu/`
- Local preview URL or route: `/en/ui-libraries/shadcn/dropdown-menu/`
- Applicable runtimes manually reviewed: not manually reviewed in the connector environment; the automated browser suite covers Web Components / React / Vue.
- [x] The existing demo consumes the real public package export and its own Trigger anatomy/state.
- [x] The Demo Matrix is supplementary evidence rather than a replacement for the website page.
- External demo orchestration: none
- Consumer-owned orchestration that is not installed with the package: none

## Validation

- Focused generated-CSS regressions: passed in CI.
- Painted browser regression: passed in Chromium for Web Components, React, and Vue in the full `test` job.
- Prototype catalog, generated style manifests, variant ordering, release tests, type contracts, and repository tests: passed.
- Workspace type check: passed.
- Public-package build and bundle budgets: passed.
- CLI and isolated React consumer smoke tests: passed.
- Release scan and release-stage dry run: passed.
- CI: workflow run #636 completed successfully across all nine jobs for head `9bf6311c1df2d9b8cefd323424219c2040508bd7`.
- Branch integrity: the branch was 0 commits behind `main` when the pull request was opened; GitHub currently reports it mergeable.
- Local limitation: a full workspace checkout was unavailable in the connector execution environment, so the authoritative full-workspace and browser evidence is the successful repository CI run above.

### DIFF
--- a/apps/www/src/content/docs/zh-cn/demo-ring-offset-default.browser.test.ts
+++ b/apps/www/src/content/docs/zh-cn/demo-ring-offset-default.browser.test.ts
@@ -1,0 +1,124 @@
+// @vitest-environment node
+
+import type { Browser, Locator, Page } from 'playwright-core';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import {
+  RUNTIMES,
+  applyColorScheme,
+  launchBrowser,
+  openRoute,
+  selectRuntime,
+  startServer,
+  stopServer,
+} from './browser-harness';
+
+const DROPDOWN_ROUTE = '/en/ui-libraries/shadcn/dropdown-menu/';
+
+let browser: Browser;
+let baseUrl = '';
+
+type RingOffsetSample = {
+  background: number[];
+  offset: number[];
+  offsetShadow: string;
+  explicitOffsetColor: string;
+  focusVisible: boolean;
+};
+
+async function focusDropdownTrigger(page: Page, previewer: Locator): Promise<void> {
+  await previewer.locator('select.adapter-select').focus();
+  await page.keyboard.press('Tab');
+  await page.waitForFunction(
+    () => {
+      const trigger = document.querySelector<HTMLElement>(
+        '[data-previewer-id] [role="button"]'
+      );
+      if (!trigger?.hasAttribute('data-focus-visible')) return false;
+      return Boolean(
+        getComputedStyle(trigger).getPropertyValue('--pui-ring-offset-shadow').trim()
+      );
+    },
+    undefined,
+    { timeout: 10_000 }
+  );
+}
+
+async function ringOffsetSample(page: Page): Promise<RingOffsetSample> {
+  return page.evaluate(() => {
+    const trigger = document.querySelector<HTMLElement>('[data-previewer-id] [role="button"]');
+    if (!trigger) throw new Error('The shadcn Dropdown Menu demo must render its trigger.');
+
+    const probe = document.createElement('span');
+    probe.style.boxShadow = 'var(--pui-ring-offset-shadow)';
+    trigger.appendChild(probe);
+
+    const triggerStyle = getComputedStyle(trigger);
+    const offsetShadow = getComputedStyle(probe).boxShadow;
+    probe.remove();
+
+    const colorMatch = offsetShadow.match(
+      /^(?:(?:rgb|rgba|hsl|hsla|hwb|lab|lch|oklab|oklch|color)\([^)]*\)|#[0-9a-f]+|[a-z]+)/i
+    );
+    if (!colorMatch) {
+      throw new Error(`Unable to read the ring offset colour from: ${offsetShadow}`);
+    }
+
+    const canvas = document.createElement('canvas');
+    canvas.width = 1;
+    canvas.height = 1;
+    const context = canvas.getContext('2d', { willReadFrequently: true });
+    if (!context) throw new Error('Canvas 2D context is required to resolve painted colours.');
+
+    const paint = (color: string): number[] => {
+      context.clearRect(0, 0, 1, 1);
+      context.fillStyle = 'rgba(0, 0, 0, 0)';
+      context.fillStyle = color;
+      context.fillRect(0, 0, 1, 1);
+      return Array.from(context.getImageData(0, 0, 1, 1).data);
+    };
+
+    return {
+      background: paint(triggerStyle.backgroundColor),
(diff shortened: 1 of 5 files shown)


guangliang2019 (approved):
Reviewed the current head against #456 and the applicable feedback-style boundary. The shared fallback now resolves implicit ring offsets through the theme background while preserving explicit offset colors, and the generated-CSS plus real Chromium coverage verifies the painted result across Web Components, React, and Vue. Scope correctly leaves #455 independent.

PR #470, opened 2026-08-22
fix(adapters): keep a not-present overlay mounted and detached

### LINKED ISSUE #392: bug(select): preserve selected textValue on first paint in React and Vue
## Summary

On current `main` (`70e2eb1a1bcd9264cf8c08e6ede210f96ee04606`), the same controlled Brutalist Select demo has different first-paint labels by runtime:

- Web Components: `Paper`
- React: `paper`
- Vue: `paper`

Opening the React or Vue Select once registers/materializes the option collection and changes the trigger to `Paper`. The selected value never changed; only the display metadata became available late.

Route: https://www.proto-ui.com/en/ui-libraries/brutalist/components/select/

## Reproduction

1. Load the route with a fresh runtime mount.
2. Select Web Components: the closed trigger reads `Paper`.
3. Select React or Vue without opening the Select: the closed trigger reads raw value `paper`.
4. Open the Select once: the trigger changes to the authored option label `Paper`.

The demo authors:

- Root value/default value: `paper`
- Item value: `paper`
- Item `textValue`: `Paper`

The first-paint output therefore leaks an internal option key in React/Vue.

## Contract

`P-BASE-SELECT-VALUE` defines Value as a renderer-consumed committed-value display and requires it to prefer Root-owned selected text before the committed raw value. `P-BRUTALIST-SELECT-VALUE` inherits that behavior.

Runtime adapter choice must not change the user-facing label or require the popup to open before selected metadata becomes usable.

## Likely cause

The React/Vue preview mounts the closed overlay without its option parts present. The selected-text metadata is therefore absent when Value first synchronizes and it falls back to the raw committed value. Opening materializes/registers items, after which the same Value renders the authored `textValue`.

This is a collection/value synchronization and adapter mount-timing problem, not a label-casing problem in the demo.

## Expected behavior

All runtimes render `Paper` on first paint, before the popup has ever opened. Selected display metadata must have a stable source that does not depend on ephemeral overlay presence.

## Proposed solution constraints

- Preserve the separation between raw committed value (`paper`) and user-facing selected text (`Paper`).
- Do not special-case this demo or title-case arbitrary values.
- Do not require `keepMounted` merely to obtain a correct closed trigger.
- Establish selected metadata from controlled props/collection registration in a lifecycle phase available consistently to Web Components, React, and Vue, or make the demo/API author the selected label explicitly through the existing contract.
- Keep collection ownership and selection ownership in their existing domains.

## Acceptance criteria

- [ ] Web Components, React, and Vue all render `Paper` on first paint with the popup closed.
- [ ] Opening/closing the popup does not change casing or replace the label unexpectedly.
- [ ] Changing the controlled value updates the corresponding authored `textValue` without a prior popup open.
- [ ] Raw values remain available to selection/event channels and are not rewritten into display labels.
- [ ] Focus, keyboard navigation, and selection behavior remain unchanged.
- [ ] Cross-runtime tests cover first paint, controlled updates, and closed-overlay remounts.

HyacinthHaru:
## Summary

Closes #392.

A closed Select showed `paper` instead of `Paper` in React and Vue until the popup had been opened once. Two defects sit behind that, and fixing only the first leaves the label wrong.

## What was actually wrong

**The closed subtree never existed.** React and Vue unmounted a not-present view, so no Item ever registered with the Root's collection and `resolveTextValue` had nothing to resolve. Web Components looked correct for a reason unrelated to view lifecycle: its Items are separate custom elements in the light DOM, so they upgrade and register whether or not the Content's view is attached. That is the reading @cyjin-yl confirmed as the basis for this work, superseding the earlier presence-parity framing.

**The resolved label never reached the screen.** With the first defect fixed, `data-display-value` became `Paper` in all three runtimes while the rendered text stayed `paper`. `base-select-value` requests a re-render only when it is already mounted:

```ts
if (requestRender && mounted) run.update();
```

In Web Components the Value mounts before the context update carrying the resolved text arrives. In React and Vue the order is reversed, so the state updated with the render suppressed, and the later `onMounted` sync returned early because the state already matched. The suppressed render is now remembered and flushed on mount.

Measured on the documentation page, closed, never opened:

| | before | after |
| --- | --- | --- |
| wc | `Paper` / `data-display-value="Paper"` | unchanged |
| react | `paper` / `data-display-value="paper"` | **`Paper` / `Paper`** |
| vue | `paper` / `data-display-value="paper"` | **`Paper` / `Paper`** |

## `data-pui-view-detached` as a contract

Per your call, the constant moved into `@proto.ui/adapter-base`, all three adapters project it, and `spec/contracts/C-HOST-VIEW-ATTACHMENT-0001.yaml` is in this PR rather than a follow-up. It sits under `C-HOST-SURFACE-PROJECTION-0001` as you suggested, as view-attachment evidence rather than a new semantic domain. Five criteria: one attribute name from one constant; present exactly when the view intent is not present; the exclusion from paint, accessibility tree, and tab order comes from the attribute's own projection; authored children still mount and register; readiness accounts for ancestry.

No `inert`, as decided. `display: none` matching Web Components is what excludes the subtree.

## Two things the change surfaced

**Scoped to presence-bound overlays.** My first attempt applied the new semantics to every view intent, which broke Tabs: an inactive panel's text stayed in `textContent`. Tabs drives presence through `lifecycle.setPresent` while the overlay families use `bindPresence`, so the overlay port now reports whether presence was bound and only those instances mount detached.

**The Vue reopen path.** Keeping the host element alive meant the same element had to initialize twice. `onMounted` called `initSession()` unconditionally, which previously could not happen while closed because no element existed; now it claimed `lastInitRoot` during the detached mount, so the real open bailed out and bound against a disposed router. `onMounted` now defers to the presence watcher.

## Verification

New browser test, `demo-select-first-paint.browser.test.ts`, asserting the label across the three runtimes with the popup never opened. It fails on unfixed `main` where it should:

```
react/label:    expected 'paper' to be 'Paper'
react/detached: expected 0 to be greater than 0
```

It deliberately does not use the shared `selectRuntime`, whose readiness waits on an exact prototype-root count — one of the things this change alters, so reusing it turned both assertions into timeouts that said nothing about the label.

Presence regressions, dialog / dropdown / select across wc, react and vue, closed then open then Escape then reopen: detached while closed, zero detached and `data-transition-state=entered` while open, detached again after Escape, `entered` again on reopen. Identical in all nine combinations. The reopen column is the path the `onMounted` guard fixes.

`vitest run packages` — 1329 passed, 34 todo, 3 skipped. `check:prototype-catalog`, `check:styles:preset`, `check:styles:variant-order`, `check:release-version`, `check:package-manifests`, `check:types:workspace` — all OK.

## Not covered

Controlled updates without a prior open are in the acceptance list and are not asserted by an automated test here; the demo has no control to drive `value` from outside. I can add a demo affordance or a fixture-driven test if you want that cell filled rather than left open.

## Source-of-truth alignment

### DIFF
--- a/apps/www/src/content/docs/zh-cn/demo-select-first-paint.browser.test.ts
+++ b/apps/www/src/content/docs/zh-cn/demo-select-first-paint.browser.test.ts
@@ -1,0 +1,138 @@
+// @vitest-environment node
+
+import type { Browser, Page } from 'playwright-core';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import type { Locator } from 'playwright-core';
+import { RUNTIMES, launchBrowser, openRoute, startServer, stopServer } from './browser-harness';
+
+const SELECT_ROUTE = '/en/ui-libraries/brutalist/components/select/';
+
+/** The demo preselects `paper`, whose authored label is `Paper`. */
+const AUTHORED_LABEL = 'Paper';
+
+/** Root, trigger, value, the closed content, and its two authored items. */
+const MOUNTED_ROOT_COUNT = 6;
+
+/**
+ * The shared `selectRuntime` waits on an exact prototype-root count, which is
+ * one of the things this file measures, so readiness here waits only for the
+ * trigger. Reusing the shared helper would turn every assertion below into a
+ * timeout that says nothing about the label.
+ */
+async function showRuntime(page: Page, previewer: Locator, runtime: string): Promise<void> {
+  await previewer.locator('select.adapter-select').selectOption(runtime);
+  await page.waitForFunction(
+    (selected) => {
+      const root = document.querySelector('[data-previewer-id]');
+      const select = root?.querySelector<HTMLSelectElement>('select.adapter-select');
+      const host = root?.querySelector('.host');
+      if (!host || select?.value !== selected) return false;
+      return Array.from(host.querySelectorAll('*')).some(
+        (element) => element.getAttribute('role') === 'combobox'
+      );
+    },
+    runtime,
+    { timeout: 20_000 }
+  );
+  await page.waitForTimeout(300);
+}
+
+let browser: Browser;
+let baseUrl = '';
+
+type ClosedSelect = {
+  label: string;
+  displayValue: string | null;
+  registeredItems: number;
+  mountedRoots: number;
+  detachedHosts: number;
+  everOpened: boolean;
+};
+
+async function readClosedSelect(page: Page): Promise<ClosedSelect> {
+  return page.evaluate(() => {
+    const host = document.querySelector('[data-previewer-id] .host');
+    if (!host) throw new Error('The Select demo must render a previewer host.');
+    const trigger = Array.from(host.querySelectorAll('*')).find(
+      (element) => element.getAttribute('role') === 'combobox'
+    );
+    if (!trigger) throw new Error('The Select demo must render a trigger.');
+    const value = Array.from(trigger.querySelectorAll('*')).find((element) =>
+      element.hasAttribute('data-pui-root')
+    );
+
+    return {
+      label: (value?.textContent ?? '').trim(),
+      displayValue: value?.getAttribute('data-display-value') ?? null,
+      // Items live inside the closed content; they register with the Root's
+      // collection even though nothing has opened it.
+      registeredItems: document.querySelectorAll('[data-collection-index]').length,
+      mountedRoots: document.querySelectorAll('[data-previewer-id] [data-pui-root]').length,
+      detachedHosts: document.querySelectorAll('[data-pui-view-detached]').length,
+      everOpened: !!document.querySelector('[data-open]'),
+    };
+  });
+}
+
+beforeAll(async () => {
+  baseUrl = await startServer(SELECT_ROUTE);
+  browser = await launchBrowser();
+}, 150_000);
+
+afterAll(async () => {
+  await browser?.close();
+  await stopServer();
+});
+
+describe.sequential('Select first paint with the popup never opened', () => {
+  it('shows the authored label in every runtime', async () => {
+    const { context, page, previewer } = await openRoute(browser, baseUrl, SELECT_ROUTE, {
+      width: 1440,
+      height: 900,
+    });
+
+    try {
+      await previewer.scrollIntoViewIfNeeded();
+      for (const runtime of RUNTIMES) {
+        await showRuntime(page, previewer, runtime);
+        const closed = await readClosedSelect(page);
+
+        // Nothing in this case may open the popup; the defect it guards only
+        // shows before the first open cycle.
+        expect(closed.everOpened, `${runtime}/never-opened`).toBe(false);
+
+        // The rendered text is the assertion that matters to a reader of the
+        // page. `data-display-value` is checked alongside it because the two
+        // drifted apart: the protocol resolved the label while the view kept
+        // painting the raw value.
+        expect(closed.label, `${runtime}/label`).toBe(AUTHORED_LABEL);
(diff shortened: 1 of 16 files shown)


cyjin-yl (commented):
Reviewed as the #392 ruling owner. This is close — the two-defect decomposition (detached subtree + suppressed-render flush), the presence-binding gate that exempts `lifecycle.setPresent` families like Tabs Content, the Vue double-init guard, and the C entity landing in-PR with sources/dependsOn/revisions all match what we agreed. The red-first evidence and the nine-combination dialog/dropdown/select presence matrix are exactly the acceptance shape.

Two things before I can approve:

1. **Controlled-updates acceptance cell is open.** In my ruling on #392 the acceptance stays *as written*, which includes controlled updates; "I can add a demo affordance or a fixture-driven test if you want" — I do want it filled in this PR rather than left open, since it is one of the four named acceptance items. A fixture-driven test (drive `value` from outside on a closed Select across the three runtimes) is enough; no demo affordance needed.
2. **`C-HOST-VIEW-ATTACHMENT-0001` has no T anchor.** Its sibling `C-HOST-SURFACE-PROJECTION-0001` carries `verifies.tests: [T-HOST-SURFACE-PROJECTION-0001]`. Criterion E (readiness accounts for ancestry) and D (children mount while detached) are both asserted by your browser tests — please add a matching `T-HOST-VIEW-ATTACHMENT-0001` that references them so the contract does not ship unverified.

Non-blocking note: criterion B says present-exactly-when-not-present — the React/Vue `detached ? "" : undefined` projections satisfy it; worth keeping an eye on any future caller tempted to reuse the attribute for pending states.

Push those two and I will approve.

cyjin-yl (approved):
增量复核通过，批准。

上轮的两个阻塞项都已按要求完成：

1. `shadcn-select-controlled-value.journey.test.ts` 在 Content 从未打开的前提下，通过真实 runtime API 连续推动两次 controlled `value` 更新，并同时断言 painted label 与 `data-display-value`；WC / React / Vue 都在同一 acceptance cell 中，且 pre-fix red 证据与原缺陷一致。
2. `T-HOST-VIEW-ATTACHMENT-0001` 现在对 A–E 五项 criterion 均有明确 case 和实现锚点；特别是 B 的双向 exactness、C 的 attribute-own projection、D 的 detached child registration、E 的 ancestry-aware readiness 都不再是正文声明而无测试关系。

我也核对了新增 contract/test 的关系方向，没有把 Tabs 这类 `lifecycle.setPresent` 家族误纳入 presence-bound overlay 语义。当前 head 的 GitHub Actions CI 已成功；Vercel 状态是外部 fork 授权失败，不构成代码阻塞。

cyjin-yl (approved):
Both cells are filled exactly as requested — approving.

1. **Controlled-updates acceptance.** The fixture-journey approach is the right call: `setup` handing the runtime API to the test models a controlled owner pushing `value` without inventing a demo control, the Content never opens (so the resolved label can only come from children that mounted while detached), two successive updates across all three runtimes with both painted label and `data-display-value` asserted. Red-first on pre-fix sources (`react/vue: 'react' ≠ 'React'`, wc green matching the original defect) plus a wait helper that names what it saw — good failure ergonomics.

2. **`T-HOST-VIEW-ATTACHMENT-0001`.** Five cases, one per criterion, each anchored to a real test — including criterion C measured at the contract layer (`adapter-base` shared rule → computed `display`) where the exclusion actually lives, and criterion B driven in both intent directions rather than closed-only. `C-HOST-VIEW-ATTACHMENT-0001` now carries `verifies.tests`, closing the shadow-guarantee concern.

The rebase onto post-#472 main is clean, and the full local suite (`1348 passed / web conformance 9 / governance green incl. public-docs + types`) matches what I expect this surface to cost. The only failing check remains the Vercel auth environment issue, not code.

Issue #458, opened 2026-08-20
spec(shadcn): distinguish pinned Button baseline from adopted current official recipe


guangliang2019:
## Summary

Issue #454 confirmed that the rendered shadcn Destructive Button can fall below a 4.5:1 text contrast ratio on Proto UI documentation surfaces. Follow-up investigation also found that the compatibility framing needs to distinguish two upstream generations:

- the pinned `P-SHADCN-BUTTON` comparison source at `shadcn-ui/ui@f31ed81983653919dd4fe77aee4b4859f610f1dc`, whose v4 new-york Button uses a solid destructive fill with white text;
- the current shadcn/ui component-and-style architecture, whose official style definitions use the same tinted destructive recipe currently projected by Proto UI:
  `bg-destructive/10 text-destructive`, `hover:bg-destructive/20`, `dark:bg-destructive/20`, and `dark:hover:bg-destructive/30`.

Maintainer direction is to preserve the official shadcn visual design rather than introduce a Proto UI-specific accessibility correction when the official recipe itself produces the observed result. The color recipe is therefore not changed by this issue.

## Scope

- Update `P-SHADCN-BUTTON`'s centralized compatibility open question to distinguish:
  - the pinned historical comparison baseline;
  - the newer official recipe adopted by the current Proto UI projection;
  - the maintainer decision to treat the measured contrast as accepted upstream-compatible behavior rather than a local palette defect.
- Add the relevant current upstream component/style sources without removing the pinned source or obscuring the historical comparison.
- Define how a fixed upstream revision and later intentionally adopted official visual recipes coexist in the compatibility record, without silently advancing guarantees.
- Record the destructive Dropdown Item branch as demonstrated with evidence or explicitly undemonstrated.
- Keep implementation, tests, and documentation claims aligned with the resulting catalog language.

## Out of scope

- Changing `--destructive` or other theme tokens.
- Replacing `text-destructive` with a Proto UI-specific foreground.
- Returning the Button to the older solid-fill recipe solely to meet a contrast threshold.
- Adding a Proto UI-specific 4.5:1 guarantee that shadcn/ui does not make for this recipe.

## Acceptance criteria

- [ ] `P-SHADCN-BUTTON-Q-UPSTREAM-DIFFERENCES` names both the pinned baseline and the adopted current official recipe, with their roles unambiguous.
- [ ] The tinted destructive recipe is classified as an intentional adoption of current upstream design, not an accidental parity gap.
[...]

PR #472, opened 2026-08-22
feat(shadcn): project Base Checkbox

### LINKED ISSUE #379: [shadcn] Project Base Checkbox into the shadcn design language
## Classification

- Ledger: shadcn design-language projection
- Base owners: `P-BASE-CHECKBOX`, `P-BASE-CHECKBOX-INDICATOR` (`status: draft`)
- New Base protocol: none

## Goal

Project the current merged Base Checkbox family into `@proto.ui/prototypes-shadcn` without changing Base checked-value, mixed-display, disabled, activation, focus, or accessibility ownership.

## Contract boundary

- Root and Indicator are the only admitted anatomy identities.
- Base owns `checked`, `indeterminate`, effective disabled state, activation, and checkbox accessibility semantics.
- The shadcn projection may own classes, token use, and design-language variants only.
- The frozen 2026-08-13 shadcn component is comparison evidence, not authority to copy unsupported API.

## Delivery surface

- `packages/prototypes/shadcn/src/checkbox/`
- `P-SHADCN-CHECKBOX` and `P-SHADCN-CHECKBOX-INDICATOR`
- exact package/root exports and public CLI registry projection
- `packages/prototypes/shadcn/test/checkbox.test.ts`
- EN/ZH docs at `apps/www/src/content/docs/{en,zh-cn}/ui-libraries/shadcn/checkbox.mdx`
- real demo `apps/www/src/content/docs/zh-cn/demo-shadcn-checkbox.demo.ts`
- preview registry, component preset/token closure, and generated Agent projection

## Acceptance criteria

- [ ] Root consumes Base Checkbox; Indicator consumes the nearest Base Checkbox context.
- [ ] The projection adds no competing value, event, focus, activation, form, or accessibility owner.
- [ ] Default, checked, indeterminate, disabled, and focus-visible presentation have executable evidence.
- [ ] Exact exports and absence of unsupported variants/events are asserted.
- [ ] Docs separate Base guarantees from shadcn visual choices and rejected upstream compatibility claims.
- [ ] The real package export renders through Web Component, React, and Vue previews.
- [ ] Focused tests, catalog, preset/token, types, docs, and proportional workspace checks pass.

## Out of scope

- Revising Base Checkbox.
- Form submission or validation.
- Toggle, Radio Group, or switch semantics.

HyacinthHaru:
Closes #379. Follows the review boundaries in https://github.com/Proto-UI/Proto-UI/issues/379#issuecomment-5376886742.

## Against the six boundaries

1. **Dark state priority.** The tint is `w.all(colorScheme dark, checked false, indeterminate false)` and lowers to `dark:not-[data-checked]:not-[data-indeterminate]:bg-input/30`. Dark + indeterminate has a computed-style regression across all three runtimes.
2. **Exact geometry.** `rounded-[4px]` and `size-3.5` with two new closures in `proto-style-css.ts`; the generated closure is updated and both are covered by a focused test.
3. **`peer` removed.** No selector in this slice consumes it.
4. **`transition-shadow` and `text-current` omitted**, both recorded as compatibility differences. `aria-invalid:*` still rejected.
5. **Absence at the right layer.** The module test asserts the Indicator has no focus handle, no event, no a11y control syntax, and no way to write what it derives. Web `tabIndex` is not written as a cross-host contract; the browser evidence asserts one focusable checkbox per instance instead.
6. **Cross-products.** Dark unchecked / checked / mixed, both-true precedence, disabled + mixed, and one-glyph-across-repeated-updates are all covered.

## Compatibility differences

Verified against the pinned `f31ed819` source rather than assumed. Intentional: `asChild`, three `aria-invalid:*`, `peer`, `text-current`, and the whole indeterminate presentation — the pinned baseline has **no mixed visual at all**, only `data-[state=checked]:*`, so mixed reusing the checked fill with a minus glyph is ours. Parity gaps: `transition-shadow`, `data-slot`, className and native prop forwarding, and the Indicator centering mechanism — upstream uses `grid place-content-center`, which is not in the closure, so this uses `flex items-center justify-center`. `size-3.5` sits on the Indicator rather than an icon inside it, because the renderer's SVG factory takes no style handle; both spellings paint a 14px glyph box.

The dark mechanism differs by necessity: upstream pairs an unguarded `dark:bg-input/30` with `dark:data-[state=checked]:bg-primary` to push the fill back on top. That re-assertion covers checked only and would miss mixed, so the guard goes on the tint instead.

## New style-compiler closures

`packages/cli/src/services/proto-style-css.ts` — `rounded-[4px]` and spacing `3.5 = 0.875rem`, both exercised by `proto-style-css.test.ts`.

A third addition was needed and is worth a separate look: `prototype-style-tokens.ts` had no `asCheckboxRoot` / `asCheckboxIndicator` entry in its per-hook state table, so **none** of the checkbox rules reached the closure — the tint landed as a plain `dark:bg-input/30` covering the filled box too. Added both entries with a regression that fails without them.

## Two things I want to flag

**The computed-style regression cannot currently fail.** I removed the guard and re-ran it: it still passes. The fill selector carries an attribute selector while an unguarded tint selector is entirely `:where()`, so the fill wins on specificity whatever the rule order. The evidence that does go red without the guard is at the token layer, in the module test and the CLI scanner test — both verified. I kept the computed-style case because it pins the painted outcome if either selector shape changes, but it is not what proves the guard.

**`w.any` does not lower.** `extractConditions` handles `eq` and `all` only, so my first version — one `any(checked, indeterminate)` rule — left the fill on the runtime plan and coloured the box a frame late. It is two same-token rules now, which is what emits `data-[checked]:*` and `data-[indeterminate]:*`.

## Platform addition outside the two closures

`SvgRootProps` gained `'aria-hidden'`, and the `svg` allowlist with it. Without it the glyph shows up as `- img` in the accessibility tree in all three runtimes, which is measured, not assumed. All three adapters already pass svg props through verbatim, so this is a type and an allowlist entry; leaf tags still reject it. Say the word if you would rather this be a separate PR and I will assert the weaker statement here instead.

## Evidence

- `packages/prototypes/shadcn/test/checkbox.test.ts`, 10 cases.
- `demo-shadcn-controls.browser.test.ts`, 2 cases over Web Components, React, and Vue.
- `vitest run packages`: 1344 passed, 34 todo, 3 skipped.
- Every governance check green: release version, release assets, style preset, variant order, component presets, prototype catalog, public docs, release tests, type contracts, agent doc, `check:types`.
- Browser suites green against one shared dev server; the checkbox route is added to `READY_ROUTES`.

Two notes on the browser suite. Teardown outgrew the default 10s hook budget once the suite went from three cases to five, so `afterAll` now declares 60s. And the Indicator glyph is a derived render that lands one turn after the Roots readiness waits on — measured at 1ms in Web Components, 3ms in Vue, 10ms in React — so the suite waits for the glyph counts before measuring rather than relying on the shared readiness helper.

`test:runtime` phase 1 fails on my machine in three `apps/www` override tests with `localStorage.clear is not a function`. That reproduces on a clean `origin/main` worktree, so it is local and not from this branch.


cyjin-yl (approved):
Reviewed against the #379 ruling boundaries I set. Approving — all six boundaries are honored:

1. **Dark priority** — `w.all(colorScheme dark, checked false, indeterminate false)` is the correct gate; dark+indeterminate computed-style regression covers the precedence trap.
2. **Geometry closures** — both new compiler closures carry focused tests.
3. **`peer` removed** — confirmed no consumer.
4. **Compatibility differences** (`transition-shadow`, `text-current`, `aria-invalid:*` rejection) — properly recorded rather than silently widened.
5. **Absence at the right layer** — module test asserting no focus handle / event syntax / ownership on the Indicator, browser evidence asserting one focusable checkbox per instance, matches the ruling exactly.
6. **Cross-products** — all six named cells covered including one-glyph-across-repeated-updates.

On the platform addition: `'aria-hidden'` on `SvgRootProps` is outside the two closures but inside the ruling's intent — the glyph must not enter the accessibility tree ("glyph 不进入可访问对象树"), the measured `- img` result justifies it, and the root-only allowlist with a leaf-rejection test keeps the widening minimal. Splitting it out would leave this PR asserting a weaker a11y statement, so keeping it here is right.

One follow-up note for the shadcn baseline bookkeeping (#458): this PR establishes the Checkbox family baseline pinning pattern — when #458 lands, the same upstream-revision treatment should be applied here for consistency.

PR #480, opened 2026-08-22
fix(style): reset composed shadow and transform properties per element

### LINKED ISSUE #455: bug(style): focus ring repaints on every descendant that carries a shadow token

HyacinthHaru:
Closes #455.

## What leaked

`composedShadow` and `transformValue` read custom properties through `var(name, fallback)`, and the fallback only applies while the property is unset. Custom properties inherit, so once an ancestor sets one, every styled descendant that composes the same shorthand paints the ancestor's value again at its own size.

Measured on clean `main`, all three runtimes, keyboard focus on the Brutalist Switch:

```
root : rgb(245,245,245) 0 0 0 2px, rgb(23,23,23) 0 0 0 4px, rgb(0,0,0) 3px 3px 0 0
thumb: rgb(245,245,245) 0 0 0 2px, rgb(23,23,23) 0 0 0 4px, rgb(0,0,0) 3px 3px 0 0
```

After the fix the thumb paints its own hard shadow and nothing else, and the root is untouched:

```
thumb: rgba(0,0,0,0) 0 0 0 0, rgba(0,0,0,0) 0 0 0 0, rgb(0,0,0) 3px 3px 0 0
```

## A second leak, same mechanism, measured

`transformValue` composes `--pui-translate-*` and `--pui-scale-*` the same way. On the Shadcn Switch, pressing the Root applies `scale-[0.98]`, and because the Root transform already scales its subtree, the Thumb applied the same factor a second time:

| | before | after |
| --- | --- | --- |
| root while pressed | `matrix(0.98, 0, 0, 0.98, 0, 0)` | unchanged |
| thumb while pressed | `matrix(0.98, 0, 0, 0.98, 0, 0)` | `matrix(1, 0, 0, 1, 0, 0)` |

The reset covers both families. I included the transform half because it is the same defect through the same helper and is fixed by the same rule — happy to split it if you would rather keep the slice to the ring.

## The fix

One `:where([data-pui-style])` rule, first inside `@layer proto-ui`, setting each composed property to `initial`.

- **Inside the layer**, because the existing box-model baseline is unlayered and unlayered styles outrank layered ones — a reset there would beat every token rule and no ring could paint at all.
- **`:where()`**, because a plain token rule is `:where(...)` too, at specificity 0. A reset with any weight would outrank it.
- **First**, so an equally weighted token rule still wins on source order.
- **`initial`** rather than literal defaults, because it restores the guaranteed-invalid value and lets the fallback already written next to each `var()` fire. Spelling the defaults a second time would be two places to keep in sync, and `--pui-ring-color` / `--pui-ring-offset-color` resolve through theme-derived fallbacks that #463 deliberately put there.

No document-wide `*` reset, per `D-WEB-STYLE-BASELINE-0001-B`.

## Evidence

New `demo-composed-style-isolation.browser.test.ts`, following the shape #463 used for the ring-offset fix. Both cases verified red before and green after:

- `wc/light/thumb-layers: expected [...] to have a length of 1 but got 3`
- `wc/pressed-thumb: expected 0.98 to be close to 1`

It resolves each shadow layer's colour through a canvas and counts painted layers, so it asserts what is drawn rather than a token string. Both cases run in Web Components, React and Vue; the ring case runs in Light and Dark.

Plus a unit test pinning the reset's placement and weight, `vitest run packages` at 1332 passed / 34 todo / 3 skipped, and every governance check green.

## Two things to flag

**Acceptance criterion 2 is wrong and I did not implement it.** I wrote "the root's hard offset shadow remains visible while focused" when filing. It cannot hold without reordering `composedShadow`, and it should not: the ring's 4px spread covers the border box in every direction, the 3px offset shadow sits inside that, and the ring is composed first so it paints on top. That order is Tailwind's, and reversing it would put every element's own shadow above its own focus ring — worse contrast on the indicator, everywhere. I suggest dropping that criterion rather than moving it; say the word if you want it as a separate design question instead.

**Criterion 3's anchors are covered by proxy, not by name.** Brutalist Button, Tabs Trigger, Toggle, Textarea and Switch all project the same `BRUTALIST_FOCUS_TOKENS` constant, so the Switch Root assertion — three painted layers, ring intact, both schemes, three runtimes — exercises the identical token set. #463's suite anchors the Shadcn side. I did not add Button and Tabs Trigger routes of their own; tell me if you want them named explicitly.

## One change outside the fix

The browser phase now runs with `--no-file-parallelism`. Adding a fifth suite made the parallel run intermittently blow `selectRuntime`'s 20s readiness timeout, because every suite drives the one shared dev server from #457 and they queue behind each other. Sequential is also faster: **75s against 102s** on five suites, with the parallel run failing three tests on the first attempt and passing on the second. Easy to drop if you would rather I fold the two cases into the existing suites instead.

Unrelated local note: `demo-brutalist-controls.browser.test.ts` times out its 10s `afterAll` under the shared-server run on my machine, sequential or parallel. It does the same on a clean `origin/main` checkout with four suites, and its `stopServer` is a no-op in that mode, so it is `browser.close()` on this hardware rather than anything from this branch. Left untouched.

### DIFF
--- a/apps/www/src/content/docs/zh-cn/demo-composed-style-isolation.browser.test.ts
+++ b/apps/www/src/content/docs/zh-cn/demo-composed-style-isolation.browser.test.ts
@@ -1,0 +1,244 @@
+// @vitest-environment node
+
+import type { Browser, Locator, Page } from 'playwright-core';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import {
+  COLOR_SCHEMES,
+  RUNTIMES,
+  applyColorScheme,
+  launchBrowser,
+  openRoute,
+  selectRuntime,
+  startServer,
+  stopServer,
+} from './browser-harness';
+
+/** Brutalist puts a hard offset shadow on the thumb, which is what re-composed the ring. */
+const BRUTALIST_SWITCH_ROUTE = '/en/ui-libraries/brutalist/components/switch/';
+/** Shadcn scales the pressed root, which is what the thumb applied a second time. */
+const SHADCN_SWITCH_ROUTE = '/en/ui-libraries/shadcn/switch/';
+
+/** `data-[pressed]:scale-[0.98]` on the Shadcn Switch Root. */
+const PRESSED_SCALE = 0.98;
+
+let browser: Browser;
+let baseUrl = '';
+
+type ShadowLayer = {
+  value: string;
+  alpha: number;
+};
+
+type SwitchShadows = {
+  focusVisible: boolean;
+  root: ShadowLayer[];
+  thumb: ShadowLayer[];
+};
+
+/**
+ * Splits a computed `box-shadow` into its layers and resolves each layer's
+ * colour through a canvas, because the themes emit oklch and no string
+ * comparison can tell a painted layer from a transparent one.
+ */
+const READ_SWITCH_SHADOWS = () => {
+  const root = document.querySelector<HTMLElement>('[data-previewer-id] [role="switch"]');
+  if (!root) throw new Error('The Switch demo must render a Root.');
+  const thumb = root.querySelector<HTMLElement>('[data-pui-style]');
+  if (!thumb) throw new Error('The Switch demo must render a styled Thumb.');
+
+  const canvas = document.createElement('canvas');
+  canvas.width = 1;
+  canvas.height = 1;
+  const context = canvas.getContext('2d', { willReadFrequently: true });
+  if (!context) throw new Error('Canvas 2D context is required to resolve painted colours.');
+
+  const layers = (element: HTMLElement) => {
+    const shadow = getComputedStyle(element).boxShadow;
+    if (shadow === 'none') return [];
+
+    const parts: string[] = [];
+    let depth = 0;
+    let current = '';
+    for (const char of shadow) {
+      if (char === '(') depth += 1;
+      if (char === ')') depth -= 1;
+      if (char === ',' && depth === 0) {
+        parts.push(current.trim());
+        current = '';
+        continue;
+      }
+      current += char;
+    }
+    parts.push(current.trim());
+
+    return parts.filter(Boolean).map((value) => {
+      const color = value.match(
+        /^(?:(?:rgb|rgba|hsl|hsla|hwb|lab|lch|oklab|oklch|color)\([^)]*\)|#[0-9a-f]+|[a-z]+)/i
+      );
+      context.clearRect(0, 0, 1, 1);
+      context.fillStyle = 'rgba(0,0,0,0)';
+      context.fillStyle = color ? color[0] : 'rgba(0,0,0,0)';
+      context.fillRect(0, 0, 1, 1);
+      return { value, alpha: context.getImageData(0, 0, 1, 1).data[3] / 255 };
+    });
+  };
+
+  return {
+    focusVisible: root.hasAttribute('data-focus-visible'),
+    root: layers(root),
+    thumb: layers(thumb),
+  };
+};
+
+async function switchShadows(page: Page): Promise<SwitchShadows> {
+  return page.evaluate(READ_SWITCH_SHADOWS);
+}
+
+/** Real keyboard focus, so the ring comes from the host's own focus-visible determination. */
+async function focusFirstSwitch(page: Page, previewer: Locator): Promise<void> {
+  await previewer.locator('select.adapter-select').focus();
+  await page.keyboard.press('Tab');
+  await page.waitForFunction(
+    () =>
+      document
+        .querySelector('[data-previewer-id] [role="switch"]')
+        ?.hasAttribute('data-focus-visible') === true,
+    undefined,
+    { timeout: 10_000 }
+  );
+}
+
+/** Horizontal scale factor of each element's own resolved transform. */
+async function switchScales(page: Page): Promise<{ root: number; thumb: number }> {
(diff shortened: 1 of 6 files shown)


cyjin-yl (dismissed):
Reviewed against #455 and the feedback-style boundary. Approving — the mechanism and the evidence both hold.

What I verified:

1. **Mechanism.** The reset writes `--pui-*` composed inputs back to `initial` (restoring the guaranteed-invalid value so the `var()` fallback stays the single default writer), scoped `:where([data-pui-style])`, first inside `@layer proto-ui` so equally-weighted token rules win on source order. The cli test pins all three ordering invariants.
2. **No cross-element reliance broken.** I checked every Brutalist hover/press translate site (`tabs/trigger.proto.ts:61`, `dropdown/trigger.proto.ts:112,117`) — the tokens are owned and consumed by the same element; nothing composes an ancestor translate/scale intentionally.
3. **Precedent-consistent packaging.** Generator change + focused test + runtime-test-plan registration without regenerating the tracked www CSS matches the merged #463 pattern; docs dev/build regenerates it.
4. **Evidence.** Three runtimes × light/dark for the focus-ring case with painted-layer counts (root 3 / thumb 1, thumb layer equal to root hard shadow), plus the pressed-scale case sampling after the transition actually moves (`DOMMatrixReadOnly` guard) rather than trusting the attribute.

One nit, non-blocking: the new browser file name says `zh-cn` but sits under the shared browser-suite location used by #470 as well — fine to leave if that is where the runner globs.

cyjin-yl (approved):
变基后的增量复核通过，重新批准。

#472 与本 PR 的冲突只发生在 `proto-style-css.test.ts` 的相邻测试插入点；当前 head 同时保留了 exact-geometry closure 用例和 composed-property reset 用例。修复本体仍满足上轮确认的三项级联条件：reset 位于 `@layer proto-ui` 内、选择器为零 specificity 的 `:where([data-pui-style])`、且出现在 token rules 之前；`initial` 继续让 shorthand 旁的 `var(..., fallback)` 成为默认值的唯一来源。

浏览器证据仍覆盖三运行时的 descendant ring isolation 与 pressed-scale isolation，GitHub Actions CI 在当前 head 成功。之前关于共享 focus token、无意依赖祖先 transform，以及 generated CSS packaging 的结论均未被这次变基改变。

PR #481, opened 2026-08-22
feat(base): give Scroll Area Viewport keyboard focus facts

HyacinthHaru:
Base prerequisite for #397, per the architecture ruling in https://github.com/Proto-UI/Proto-UI/issues/397#issuecomment-5376888021. The Brutalist projection follows in a second PR; #397 stays open until that one lands.

## Against the five points

1. **Focus lands on Base.** `setupScrollAreaViewport` composes the existing `asFocusable()` and exposes `focused` and `focusVisible`. It is the only Scroll Area part that installs a focus domain, so the styled layer only reads facts.
2. **Participation follows the scrollable directions.** `none` at setup, because geometry has not arrived and a surface with nowhere to scroll is an empty tab stop; then `auto` while any of the four `canScroll*` facts is true and `none` when all are false. Losing the last direction calls `setNavParticipation` only, which rewrites the tab stop without touching the focus owner — asserted directly.
3. **No new identities.** Only `C-AS-FOCUSABLE-0001` is added to `dependsOn`. `P-BASE-SCROLL-AREA-VIEWPORT` gains two criteria and `T-BASE-SCROLL-AREA-0001` one case; the a11y-neutrality criterion now says taking focus is not a reason to invent a role either.
4. **Nothing else moved.** No ARIA role or name, no second host wrapper, no change to scroll request or fact ownership, no DOM target exposed, no Brutalist token.
5. **Regressions** below.

I also did not publish a state named `focusable`. Scrollability is read from the four direction facts and never surfaces under that name.

## Evidence

Module tests in `packages/prototypes/base/test/scroll-area.test.ts`, five new cases, each verified red before:

| case | red without the slice |
| --- | --- |
| scrollable Viewport gets focus facts, no widget role | `Cannot read properties of undefined (reading 'get')` |
| nothing to scroll stays out of sequential navigation | `expected null to be '-1'` |
| follows content between scrollable and not | `expected null to be '-1'` |
| focus stays put when the last direction goes away | `Cannot read properties of undefined (reading 'get')` |
| keyboard entry versus pointer entry | `Cannot read properties of undefined (reading 'get')` |

Browser regressions in `demo-base-controls.browser.test.ts`, four new cases over Web Components, React and Vue:

- a tab stop only on the Viewport that can scroll, `0` against `-1`, no role on either, and exactly one focusable node inside a scrolling Root and none inside a fitting one;
- content driven from nothing to scroll, to scrollable, and back, asserting the tab stop each way;
- keyboard entry giving `focused` and `focusVisible`, pointer entry giving `focused` alone;
- Arrow, PageDown, End, Home and wheel scrolling still moving the focused Viewport.

The first three fail on the assertion that names the defect rather than on a readiness timeout: `wc/scrolling-tabindex: expected null to be '0'`, `wc/fitting-before: expected null to be '-1'`, `wc/keyboard-focused: expected false to be true`. Readiness waits on the scroll fact, which is published with or without this slice, which is what makes that possible. The fourth passes on `main` too — it is a non-regression guard, not a witness.

`vitest run packages`: 1336 passed, 34 todo, 3 skipped. Every governance check green, plus `check:types`.

## Two things in the demo

The Base Scroll Area demo gains a second surface that fits its content, because the no-overflow half of the acceptance list needs one to exist.

It also needed a correction to be evidence at all: its rows were plain strings in one box, so the renderer concatenated them into a single 97px text node inside a 190px viewport. **The demo has never scrolled on `main`.** Each row is its own box now, and the scrolling surface actually overflows. Screenshot-free but measurable: `clientHeight` 190 against `scrollHeight` 190 before, and the vertical direction facts were absent.

## One observation, not addressed here

`demo-base-controls.browser.test.ts` goes from one case to five. Under four browser suites running in parallel against the one shared dev server, its `afterAll` then hangs on my machine — 60s is not enough, so it is a hang rather than slowness, and the untouched Brutalist suite does the same. Running the suites sequentially clears it: all fourteen tests and both teardowns pass. That contention is what #480 addresses by making the browser phase sequential, so I left this suite alone rather than papering over it here. If CI shows it on this branch I will rebase onto #480 or carry the flag here.

### DIFF
--- a/apps/www/src/content/docs/zh-cn/demo-base-controls.browser.test.ts
+++ b/apps/www/src/content/docs/zh-cn/demo-base-controls.browser.test.ts
@@ -19,6 +19,12 @@ const COLOR_SCHEMES = ['light', 'dark'] as const;
 type ColorScheme = (typeof COLOR_SCHEMES)[number];
 
 const TEXTAREA_ROUTE = '/en/ui-libraries/base/textarea/';
+const SCROLL_AREA_ROUTE = '/en/ui-libraries/base/scroll-area/';
+
+/** The demo order: one surface that overflows, one that fits its content. */
+const SCROLLING = 0;
+const FITTING = 1;
+const VIEWPORT_SELECTOR = '[data-pui-scroll-projection]';
 const TEXTAREA_DEMO_SCOPE = '[data-demo-id="demo-base-textarea"]';
 const TEXTAREA_OUTPUT_SURFACES = ['stateLabel', 'eventLog', 'help'] as const;
 const MIN_TEXT_CONTRAST = 4.5;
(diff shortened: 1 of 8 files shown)


cyjin-yl (approved):
Reviewed as the Base Scroll Area domain spot-checker and against `P-BASE-SCROLL-AREA-VIEWPORT`. Approving.

What I verified:

1. **Focus lands on the scrolling surface, nowhere else.** The Viewport is the only anatomy part installing a focus domain; Root stays behavior-free (`ROOT-NO-BEHAVIOR` intact — I browser-probed exactly this boundary in an earlier audit: no tabindex, no exposes on Root). The styled layer consuming facts only matches the Brutalist projection split you announced for the follow-up PR.
2. **Participation semantics are right and well-guarded.** `none` before geometry, `auto` iff any of the four direction facts is true, `setNavParticipation` only (no blur path), and the explicit refusal to publish a state named `focusable` — that last one matters because the Focus module already owns that name for a different fact.
3. **Spec projection is coherent.** Two new criteria with `C-AS-FOCUSABLE-0001` dependsOn, the a11y-neutrality criterion extended to cover "takes focus ≠ invents a role", revision entry at 0.3.0-alpha.0, T case added — the source-of-truth loop guangliang2019 has been enforcing elsewhere is closed here.
4. **Evidence quality.** Five red-first module cases plus a real-keyboard Tab browser regression separating keyboard entry from pointer entry across all three runtimes — this directly supersedes the "viewport not tab-reachable" observation from my earlier spot-check on `main`, which was the gap this PR closes.

One forward note, non-blocking: when the Brutalist follow-up lands, its P entity should consume `focused`/`focusVisible` for any focus-visible ring rather than re-declaring focus protocol — same consumption-only rule the criteria now state.

PR #482, opened 2026-08-22
fix(brutalist): ring the focused Scroll Area from inside

HyacinthHaru:
Second half of #397, per https://github.com/Proto-UI/Proto-UI/issues/397#issuecomment-5376888021. #397 closes with this one.

**Depends on #481.** A cross-fork PR cannot use another fork branch as its base, so this one targets `main` and carries #481's four commits underneath. The three that belong here are:

- `fix(cli): lower Scroll Area Viewport focus conditions`
- `fix(brutalist): ring the focused Scroll Area from inside`
- `test(docs): cover the Scroll Area focus ring in three runtimes`

Merge #481 first and this collapses to those three on its own.

## The projection

`focusVisible` comes from the Base Viewport that #481 gives it, and the rule paints `ring-2 ring-ring ring-inset ring-offset-0`. The shared `BRUTALIST_FOCUS_TOKENS` are deliberately not reused: they draw outward and the Root's `overflow-hidden` clips them, which is why a keyboard user never saw which surface was active. The Viewport also gains `outline-none`, so the UA outline the issue measured at 1.06:1 on the Dark Root stops being the only indicator.

No second focus domain here; this layer only reads the fact.

## One thing the projection needed first

The rule did not lower. `prototype-style-tokens.ts` keys variants off a per-hook state table with no `asScrollAreaViewport` entry, so `ring-2 ring-inset` reached the closure **unconditional** — the ring would have painted at rest. Added the entry with `focused` and `focusVisible`, plus a scanner regression. The generated Brutalist closure gains `data-[focus-visible]:ring-inset` and `data-[focus-visible]:ring-offset-0`.

This is the same missing-registration shape I hit in #472 for the Checkbox hooks. Two components in a row suggests the table is worth a look as a mechanism, but I left that alone here.

## Evidence

Module test in `packages/prototypes/brutalist/test/scroll-area.test.ts`: the four inset tokens lower, `ring-offset-2` does not, no ring at rest, and the ring appears from the inherited fact after keyboard focus.

Browser regression in `demo-brutalist-controls.browser.test.ts`, over Web Components, React and Vue, in Light and Dark:

- exactly one inset layer, matching `0px 0px 0px <positive>px inset` — zero offsets with a positive spread is what "visible on all four sides" is in computed-style terms;
- every other composed layer draws nothing, so there is no outward ring for the Root to clip;
- Viewport bounds compared at rest, focused, and after scrolling, so the ring is not paid for with a geometry shift;
- both axes scrolled while the ring is up, waiting on the position rather than a fixed delay.

Verified red by swapping the rule back to the shared outward tokens: `wc/light/inset-count: expected [] to have a length of 1 but got +0`.

The demo gains a row wider than the Viewport and a horizontal Scrollbar, because the acceptance list asks for both axes and it only scrolled vertically. Refs are new names, so the existing 320px containment case is untouched.

`vitest run packages`: 1339 passed, 34 todo, 3 skipped. Web conformance green. Every governance check green, plus `check:types`. All 15 browser tests pass against one shared dev server.

## Not covered, and why

Drag geometry is asserted as "bounds unchanged", not as a drag journey. `packages/web-conformance/test/brutalist-scroll-area-move.journey.test.ts` already owns Thumb dragging and stays green; adding a second drag path here would duplicate it. Say the word if you want the ring asserted mid-drag as well.

The `afterAll` teardown of the Base and Brutalist browser suites intermittently hangs on my machine under the shared-server run — 60s is not enough, so it is a hang, and the Brutalist one does it on a clean `origin/main` checkout too. All 15 tests themselves pass. Noted rather than papered over; #480 reduces the contention that triggers it.


cyjin-yl (approved):
按 #481 之上的三提交增量审查，批准；合并顺序仍应严格为 #481 → #482。

核对结果：

1. Brutalist Viewport 只从 `asScrollAreaViewport().stateHandles` 消费 `focusVisible`，没有安装第二个 Focus domain，也没有扩宽 styled-layer 的公开 expose surface。
2. `asScrollAreaViewport` 的 lowering 表同时登记 `focused` / `focusVisible`，scanner 回归能防止条件 token 被误生成为无条件 token；生成闭包中的 ring 规则保持 state-gated。
3. 选择 `ring-2 ring-ring ring-inset ring-offset-0` 正确避开 Root 的 `overflow-hidden`，且没有复用会向外绘制的共享 Brutalist focus tokens。
4. 浏览器证据覆盖 WC / React / Vue × Light / Dark，断言恰好一个 inset layer、无可绘制的 outward layer、焦点前后及滚动后 bounds 不变，并在 ring 存在时继续推动纵横两轴。
5. `outline-none` 在本仓库生成的是 `outline: 2px solid transparent` 而非彻底移除 outline，因而仍保留 forced-colors 下由系统色替代的 fallback。
6. 当前 head 的 GitHub Actions CI 成功；#434 已完成同一 issue 的 thumb/track 对比度部分，因此 #482 在 #481 之后合并即可关闭 #397。

由于 #482 暂时携带 #481 的提交，#481 合并后的 rebase 若触发 GitHub dismiss approval，只需按独立三提交重新确认即可；当前内容本身没有阻塞项。

Issue #377, opened 2026-08-13
[tracking] Maintain the v0.3 Prototype Coverage Matrix


cyjin-yl:
## Goal

Maintain the authoritative 2026-08-13 v0.3 disposition of all 64 frozen shadcn directory entries plus Proto UI-owned Base and Brutalist candidates.

Roadmap source and frozen comparison baseline: PR #375. Stable semantics remain governed by `spec/**`. This Issue tracks classification and dependencies; it is not an implementation contract and does not require every classified entry to ship in 0.3.

## Completion and reporting rules

Every frozen entry has one primary identity from `existing`, `semantic`, `projection`, `styled-only`, `composition`, `provider/system`, or `excluded`; explicit Base, shadcn, and Brutalist dispositions; dependencies; owner; difficulty; wave; state; and a linked Issue after the bounded work item has an explicit authority or decision state plus an evidence and validation boundary. Creating or updating a reversible collaboration Issue under current authorization is not a privileged publication gate.

Classification closure is not implementation closure. Report `implemented`, `ready`, `research`, `blocked`, `deferred`, `composition/provider/system`, and `excluded` separately. Counts such as 51, 63/64, and Base 30–35 are observations, never implementation or release gates.

Current frozen routing counts:

- identities: 8 `existing`, 26 `semantic`, 5 `projection`, 7 `styled-only`, 16 `composition`, 1 `provider/system`, 1 `excluded`;
- states: 8 `implemented`, 6 `ready`, 11 `research`, 12 `blocked`, 9 `deferred`, 17 `composition/provider/system`, 1 `excluded`.

A row's state describes its current roadmap action. A design-language-specific implementation already present in another column does not silently change a missing shadcn disposition to implemented.

## Coverage Matrix

| Entry | Snapshot group | Primary identity | Base | shadcn | Brutalist | Dependencies | Owner | Difficulty | Wave | State | Issue/PR |
|---|---|---|---|---|---|---|---|---|---|---|---|
| Button | Existing | `existing` | P-BASE-BUTTON (merged) | implemented projection | implemented semantic projection | merged Base P/T; no blocker | maintainers | snapshot | existing | `implemented` | — |
| Dialog | Existing | `existing` | P-BASE-DIALOG* (merged) | implemented projection | implemented semantic projection | merged Base P/T; no blocker | maintainers | snapshot | existing | `implemented` | — |
[...]

PR #497, opened 2026-08-24
docs(records): Select item-aligned positioning design proposal

cyjin-yl:
Adds the design-only, non-normative proposal drafted under the #495 maintainer checkpoint (2026-08-24), mirrored as bounded proposal issue #496.

Contents: information path, the four checkpoint focus areas (selected Item target lifetime; scroll/measure/place/collision ordering with two falsifiable strategies; async mount and open-time selection changes; host fallback), negative boundaries, cross-host assumptions, falsification conditions, and open implementation questions reserved for checkpoint.

No implementation, spec entity admission, owner decision, or #377 rescheduling is authorized or proposed by this record.


guangliang2019 (commented):
No actionable technical finding on current head `e234c2bf207fcd185458328eb6ce8731ea0db71b` for the design-only record.

The record remains non-normative, preserves host geometry and selection ownership, and correctly leaves upstream/carrier facts as unknowns. This `COMMENT` is not semantic acceptance of strategy A/B, carrier shape, collision fallback, open-time re-placement, or implementation authorization; those decisions remain on #496. No executable tests were expected for this record-only change, and live upstream behavior was not independently verified.

guangliang2019 (approved):
Approved on exact head e234c2bf for the non-normative design record only. This approval does not admit strategy A/B, carrier shape, collision fallback, open-time replacement, an entity, or implementation; those decisions remain on #496.

PR #499, opened 2026-08-24
docs(records): Switch constrained drag-to-value design proposal

cyjin-yl:
Adds the design-only, non-normative proposal drafted under the #495 maintainer checkpoint, mirrored as bounded proposal issue #498.

Contents: information path over HC-MOVE-GESTURE-0001, responses to all six checkpoint focus areas (press/tap arbitration; activation threshold; provisional host-local position via a neutral style-projection seam; commit/cancel across all five cancel reasons; controlled truth rollback without optimistic local writes; single activation invariant), keyboard/a11y invariants, carrier-shape candidate (Switch-domain asHook, no generic drag-to-value API), cross-host fallback, falsification conditions including the Move-substrate sufficiency test, and calibration items reserved for implementation.

No implementation, public API, entity admission, owner decision, or #377 rescheduling is authorized or proposed by this record.

### DIFF
--- a/internal/records/2026-08-24-switch-drag-to-value-proposal.zh-CN.md
+++ b/internal/records/2026-08-24-switch-drag-to-value-proposal.zh-CN.md
@@ -1,0 +1,105 @@
+# Switch constrained drag-to-value proposal（design-only，non-normative）
+
+日期：2026-08-24
+
+状态：design proposal 草案。本文是短期方向记录，不是 Proto UI 语义来源；不含任何已授权的实现、实体准入或 owner 定案。授权来源：#495 maintainer checkpoint（2026-08-24，guangliang2019）：Switch 作为 `C-MOVE-GESTURE-0001` 的第二消费者验证载体，领域专项设计，不产出通用 `asMove()` / `asDrag()` / drag-to-value API。镜像有界 proposal issue：#498。
+
+## 问题与定位
+
+当前 Switch 只支持点击翻转：root 经 `press.commit` 一次性取反 checked（`packages/prototypes/base/src/switch/root.proto.ts`），thumb 是二态 CSS 切换（如 shadcn thumb 的 `translate-x-0` ↔ `translate-x-[calc(100%_-_2px)]`）。拖动时 thumb 不跟手。
+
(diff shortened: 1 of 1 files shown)


guangliang2019 (dismissed):
Requesting changes on current head `f078a0825542c03afa32a41468bb8741d89bdce4`.

**[P2 · PUI-PR499-THRESHOLD-SUBSTRATE-001] The proposal binds a carrier shape that the current Move substrate does not provide.** `internal/records/2026-08-24-switch-drag-to-value-proposal.zh-CN.md:25` assumes `activation: threshold` and an unchanged tap path below threshold. Current `MoveGestureActivation` is `immediate` only; the Web host accepts on `pointerdown`, prevents default, captures, and emits `start` immediately. Threshold and press arbitration are not stable guarantees.

Please replace the unsupported binding with an explicit decision gap: either define how an immediate Move session composes with `asTrigger` while preserving press, or obtain a separate D/C/HC/T-MOVE threshold/arbitration extension before authorizing Switch implementation. This review does not choose the threshold, paint seam, or public carrier; those remain maintainer design decisions.

guangliang2019 (commented):
Fresh partial re-review of current head `439fd51a83f9285a2f67d8ec304ad8ce2d4b44c0`.

Reconciliation:
- `PUI-PR499-THRESHOLD-SUBSTRATE-001` — resolved. The record now states the current immediate-only Move behavior (pointerdown acceptance, default-action cancellation, capture, and start) and makes Switch implementation depend on a separately approved choice between immediate Move/`asTrigger` arbitration and a D/C/HC/T-MOVE pending-threshold extension.
- No new blocking findings in the examined record-only scope.

The revised proposal keeps the important boundaries explicit: Root remains the checked/activation/commit owner; Thumb gains no value/event/focus/a11y ownership; samples and provisional geometry remain host-local; threshold, velocity, paint carrier, exact geometry target, and public/asHook shape remain candidates; implementation, entity admission, public API, owner finalization, and #377 scheduling remain gated.

Evidence for this exact head: live CI type-check/test/DCO and applicable checks succeeded; `git diff --check` passed in a fresh `/tmp` clone; all named spec/test/implementation paths resolve. Release-only jobs are skipped and do not contribute semantic acceptance.

Limitation: this is a non-dispositive partial review (no fresh local self-assessment was available for a full cross-domain review), so it is submitted as COMMENT, not APPROVE. Cross-host feasibility was not exercised against a non-Web adapter and remains a maintainer design question.

guangliang2019 (approved):
Approved on exact head 439fd51a for the non-normative design record. The unsupported threshold/substrate assumption is removed. This approval does not choose the A/B substrate direction or authorize Switch implementation; that decision remains on #498.

PR #484, opened 2026-08-22
fix(event): transactional listener installation and portable default-action control

cyjin-yl:
## Summary

Implements **PR 1** from the #466 third-pass investigation, which the thread identified as the top code priority before any GPUI renderer work.

## What changed

1. **Transactional listener installation.** `EventKernel.bindAll()` now rolls back every earlier attachment when a later addEventListener throws, leaving all registrations logically unbound so an explicit retry installs the complete plan. Previously only target *resolution* was preflighted; a mid-plan attachment failure leaked earlier listeners.
2. **Portable immutable data-only payload.** Event boundary constructs a fresh immutable view for every portable callback. Default-action control (`ev.control`) is live only during the synchronous callback window and fails closed afterward. Multiple registrations for one host sample are deduplicated. Raw native `preventDefault`/`stopPropagation` are never available on portable payloads.
3. **Semantic/host options boundary.** Semantic (non-`host:*`) registrations reject listener options at setup. `@proto.ui/types` no longer depends on DOM `AddEventListenerOptions`.
4. **Base Button migration.** The focused-Space handler now requests prevention via `ev.control`; the contract test asserts through the `EVENT_CANCEL_DEFAULT_ACTION_CAP` collector.
5. **Spec closure.** M-EVENT-0001-J defines transactional rollback, idempotent cleanup, window-bounded control, sample deduplication, and semantic options rejection. T-EVENT-0003 adds three cases.

## Exclusions / residual risk

- No broader Event routing, opaque routes, Focus, or GPUI admission change.
- The immutable payload contract applies only to portable (semantic) events; `host:*` extensions keep raw host events.
- Merge-time tests cannot prove the full default-branch E2E, live OAuth, or failure convergence; these remain post-merge production gates.

## Validation (current head `6f76ed60`)

- Event facade/impl/contract + Adapter Base default-action + Base Button: 4 files / 39 tests passed.
- Event/related broader validation: 87 files / 600 tests passed; browser journeys 19/19.
- `check:types` — 0 errors.
- `check:prototype-catalog` — OK.
- Local actionlint was not independently run; the existing CI lanes remain the required evidence.

## Contribution provenance

- **Original rights:** The Event kernel changes, immutable payload construction, options boundary, Base Button migration, and spec/test updates are original work created for Proto UI. The DCO signer confirms the right to submit them under the repository MIT License.
- **Third-party/upstream:** No code or assets were copied or ported; the implementation relies only on the existing Proto UI Event/Button contracts and public DOM event behavior.
[...]

PR #511, opened 2026-08-25
fix(radio-group): migrate stale ev.detail to portable payload after #484 merge

cyjin-yl:
## Summary

Fix a post-merge regression on `main` caused by #484's portable payload migration: `packages/prototypes/base/src/radio-group/item.proto.ts` still had a stale `getKeyboardEvent` helper reading `ev.detail` and a stale `key.down` handler using `event?.detail?.key`. After #484 migrated all other base prototypes to the immutable data-only `ProtoEventPayload`, these two stale references caused `check:types` and `test` failures on every PR rebased onto current `main`.

## Changes

- Remove the stale `getKeyboardEvent` helper (reads `ev.detail` / `ev.nativeEvent`, neither exists on the portable payload).
- The Space handler already uses `event.control.requestDefaultActionPrevention` (fixed in the same file in a prior commit); this removes only the stale Enter-check path.
- The Enter check now reads `event?.key === 'Enter'` directly, matching the portable payload contract.

## Validation

- `check:types` — 0 errors (168 files).
- Radio Group focused: Base 12/12, React 1/1, Vue 1/1, WC 1/1 = 15/15.
- `test:runtime` — 20/20.
- `check:prototype-catalog` — OK.
- `test:public-docs` — passed.

## Contribution provenance

- **Original rights:** This is a corrective migration of existing Proto UI code to match the portable payload contract introduced by #484. The DCO signer confirms the right to submit it under MIT.
- **Third-party/upstream:** No code or assets were copied or ported.
- **Material AI assistance:** An AI coding assistant (OpenAI Codex/GPT-5.6 family via Oh My Pi) materially assisted with the regression diagnosis and fix. The human contributor reviewed the diff and validated the type checks, focused tests, runtime suite, catalog, and public docs. No third-party or private code was supplied to the model.
- **Employer/client:** The signer confirms no employer/client proprietary code is included and that they have the right to publish this contribution.

### DIFF
--- a/packages/prototypes/base/src/radio-group/item.proto.ts
+++ b/packages/prototypes/base/src/radio-group/item.proto.ts
@@ -16,14 +16,6 @@ import type {
   RadioGroupItemProps,
 } from './types';
 
-function getKeyboardEvent(event: unknown): KeyboardEvent | null {
-  const detail = (event as CustomEvent | undefined)?.detail ?? event;
-  if (detail instanceof KeyboardEvent) return detail;
-  if (!detail || typeof detail !== 'object') return null;
-  const native = (detail as { nativeEvent?: unknown }).nativeEvent;
-  return native instanceof KeyboardEvent ? native : null;
-}
-
 function setupRadioGroupItem(def: DefHandle<RadioGroupItemProps, RadioGroupItemExposes>): void {
   asTrigger();
   const focusable = asFocusable<RadioGroupItemProps>();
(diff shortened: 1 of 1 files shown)


guangliang2019 (approved):
Approved on exact head `a55388b64223c8af97fa519e1dc4b26c5b365b93`.

This is a bounded post-#484 compatibility repair: the Radio Group client now reads the portable payload instead of stale raw `ev.detail`. The diff is limited to the owning client surface and introduces no new semantic or public boundary.

Independent validation covered Radio Group 12/12, Event/Adapter 39/39, type checking, the prototype catalog, and all 13 live current-head checks including DCO. No review threads or findings remain.

This approval does not authorize merge, release, or any scope beyond the isolated portable-payload migration.

PR #488, opened 2026-08-22
spec(shadcn): record adopted Button recipe separately from pinned baseline

cyjin-yl:
## Summary

Refs #458.

This PR makes the compatibility record match the maintainer decision without changing the Button implementation or visual recipe:

- keeps `shadcn-ui/ui@f31ed81983653919dd4fe77aee4b4859f610f1dc` as the pinned historical compatibility comparison baseline;
- records the current tinted destructive Button recipe as an intentional adoption of the later official shadcn/ui recipe at `1773ecfeeb4a04366978d353e69b5c7ded78dcb2` (`apps/v4/registry/styles/style-maia.css`), explicitly without advancing the overall compatibility baseline;
- retains #454 contrast measurements as adopted-upstream behavior evidence rather than an accidental Proto UI parity gap, local palette remediation requirement, or Proto UI-owned 4.5:1 guarantee;
- records destructive Dropdown Item as `projection-covered / render-undemonstrated`: the existing unit test asserts the `text-destructive` projection, while the docs surface has no corresponding rendered evidence;
- updates `T-SHADCN-BUTTON-0001` evidence wording and the 0.3 revision history to reflect the same boundary.

## Scope boundary

This is catalog/evidence consistency only. It does not change `packages/prototypes/shadcn/src/button/**`, add a destructive docs demo, advance the pinned compatibility baseline, or create new P/T identities.

## Evidence

- Current implementation uses the tinted destructive tokens documented by #458.
- `packages/prototypes/shadcn/test/dropdown.test.ts` constructs a destructive item and asserts `text-destructive`, which is projection evidence only.
- The pinned upstream source remains the v4 new-york Button at `f31ed819...`.
- The later exact official style source at `1773ecfe.../apps/v4/registry/styles/style-maia.css` contains the destructive recipe `bg-destructive/10`, `hover:bg-destructive/20`, `dark:bg-destructive/20`, `text-destructive`, and `dark:hover:bg-destructive/30`.

## Validation

- Static YAML/source review completed against current `main` (`45d19a8b23e195cf76a92f52c86ccbeb5e6486f3`).
- Local repository commands were not run in this execution environment because it could not resolve GitHub to obtain a checkout; required repository CI must therefore be the executable validation gate.

Signed-off-by: cyjin-yl <cyjin-yl@users.noreply.github.com>

Issue #454, opened 2026-08-19
bug(shadcn): destructive Button reads below 4.5:1 in both themes


HyacinthHaru:
## Summary

The shadcn Destructive Button reads below the 4.5:1 body-text floor in both themes. Unlike the Brutalist case in #449 this is not a mistyped token — it is a deviation from the pinned upstream recipe, so it belongs in the compatibility conversation rather than in the palette.

Baseline: `d7852bad`

Route: https://www.proto-ui.com/en/ui-libraries/shadcn/button/

## Browser evidence

Resting Destructive button, colours resolved through a canvas and composited to the first opaque ancestor, Web Components:

| theme | text | effective background | contrast |
| --- | --- | --- | --- |
| light | `rgb(231, 0, 11)` | `rgb(243, 220, 221)` | **3.66:1** |
| dark | `rgb(255, 100, 103)` | `rgb(80, 49, 50)` | **3.98:1** |

| light | dark |
| --- | --- |
| ![light](https://raw.githubusercontent.com/HyacinthHaru/Proto-UI/audit-artifacts/2026-08-18-visual-audit/shadcn-destructive-light.png) | ![dark](https://raw.githubusercontent.com/HyacinthHaru/Proto-UI/audit-artifacts/2026-08-18-visual-audit/shadcn-destructive-dark.png) |

## Why this is a compatibility question, not a token fix

`packages/prototypes/shadcn/src/button/button.proto.ts:35`:

```
destructive: 'border-transparent bg-destructive/10 text-destructive'
```

The pinned baseline in `P-SHADCN-BUTTON` is the v4 new-york Button at `f31ed81983653919dd4fe77aee4b4859f610f1dc`, whose destructive variant is a **solid fill with white text**. Ours is a tinted fill with the destructive colour as text. That is a design change, and it is the reason the pair lands where it does.

**Reverting to the upstream recipe is not automatically the answer**, which is why I am filing rather than fixing. Taking the same `--destructive` values this projection already resolves:

| recipe | light | dark |
| --- | --- | --- |
| ours — tinted fill, destructive text | 3.66:1 | 3.98:1 |
| upstream — solid fill, white text | **4.77:1** | **2.89:1** |

So upstream's own recipe clears the floor on Light and falls further below it on Dark. Whatever is chosen has to be chosen with both themes in view.

[...]

PR #583, opened 2026-09-01
spec(shadcn): record the adopted Checkbox recipe apart from the pinned baseline

HyacinthHaru:
Applies the #458 treatment to the Checkbox family, the follow-up cyjin-yl left on #472:

> this PR establishes the Checkbox family baseline pinning pattern — when #458 lands, the same upstream-revision treatment should be applied here for consistency.

Same shape as #488 for Button: the pinned historical compatibility baseline and evidence for later official recipes are recorded separately, and adoption is stated per item.

## What the measurement turned up

`apps/v4/registry/new-york-v4/ui/checkbox.tsx` is **byte-identical** between the pinned `f31ed819` and current upstream `main`, so unlike Button there is no newer generation of the component file. The later official Checkbox visual lives elsewhere — `.cn-checkbox` and `.cn-checkbox-indicator` in `apps/v4/registry/styles/style-maia.css` at `1773ecfe`, the same revision #488 pinned for Button.

Comparing it item by item changes one of my own entries:

| later official recipe | this projection | disposition |
| --- | --- | --- |
| `flex items-center justify-center` | same | **not a parity gap** — recorded as one in #472, retracted here |
| `[&>svg]:size-3.5` | 14px glyph box | agrees |
| `rounded-[6px]` | `rounded-[4px]` | **not adopted**; the radius follows the pinned baseline per your #379 ruling |

The centering entry is the substantive correction. #472's Indicator question said we use `flex items-center justify-center` "rather than `grid place-content-center`" because the latter is outside the closure, and filed the difference as a gap against upstream. The later official recipe centers with flex too, so what we do is the current official spelling and the `grid` in the pinned file is the older one. That also retires the open sub-question of whether `grid place-content-center` is worth a closure extension — it is not.

This is the same failure mode as #454, where I read the pin as "what upstream does" and framed our recipe as the deviation. The distinction #488 introduced is what makes it visible, which is why applying it here was worth doing rather than a bookkeeping chore.

## Changes

- `P-SHADCN-CHECKBOX` — statement, `COMPATIBILITY-SUBSET` criterion, open-question context, relabelled pinned source, new maia source with a non-advancing label, revision entry.
- `P-SHADCN-CHECKBOX-INDICATOR` — centering recorded as agreement rather than a gap, closure sub-question dropped, same source and revision treatment.
- `packages/spec/graph/test/shadcn-checkbox-baseline.test.ts` — a loader test in the shape of the one #488 added, asserting the **parsed** bilingual text. It pins both halves of the criterion (agreement is not baseline movement, disagreement is not automatically a gap), the three per-item conclusions, and the Indicator retraction, so an edit that drops or truncates any of them fails.

No source, token, or behaviour change; the passing guarantee is untouched.

`vitest run packages` 1481 passed / 34 todo / 3 skipped, and the catalog, agent-doc, public-docs and style-preset checks are green.

One note from syncing: this branch is on today's `main`, which needed a `pnpm install` for the new `vue2` adapter workspace before its suites resolve `vue`. Nothing to do in this PR, just in case anyone else hits 20 red files after pulling.


cyjin-yl (changes requested):
Reviewing exact head `666a864cea75b6ee8271c135a2b150e47b53bd9f`.

I’m requesting changes for two blocking spec/evidence inconsistencies:

1. `spec/prototypes/P-SHADCN-CHECKBOX.yaml` now records that `apps/v4/registry/new-york-v4/ui/checkbox.tsx` is byte-identical between the pinned historical revision and “current upstream main”, and the new graph test deliberately preserves that floating wording. I independently checked shadcn/ui `main` at `63c1308d112b6b1205d86244a156cca1abef5087`; its `checkbox.tsx` is indeed byte-identical to the pinned `f31ed819...` file today. The problem is reproducibility: the entity’s `sources` list does not pin the exact compared `checkbox.tsx` revision, so once upstream `main` moves this catalog claim can no longer be reconstructed from its own evidence. Please replace the floating comparison with the exact measured SHA (or another exact compared revision), add that revision’s `checkbox.tsx` as a source, and make the focused test preserve the pinned evidence rather than the phrase “current upstream”.

2. The revised `P-SHADCN-CHECKBOX-COMPATIBILITY-SUBSET` criterion requires every difference from the pinned baseline to be classified as an intentional deviation or parity gap. The same open-question context now says flex centering agrees with the later official recipe, and `P-SHADCN-CHECKBOX-INDICATOR` says the centering mechanism is no longer a parity gap / the `grid place-content-center` closure is unnecessary; however, the Root context’s final remaining-gap sentence still lists “the Indicator centering mechanism” as an implementation/review gap. Please settle this classification consistently against the pinned baseline (for example, explicit intentional adoption of the later recipe if that is the intended conclusion) and remove the contradictory remaining-gap entry, or keep it as a gap consistently. The focused test should assert the settled classification rather than only the retraction wording.

Exact-head repository CI is green (`CI` run 33512055824, including `test`, `type-check`, `lockstep-check`, and `release-scan`) and DCO reports all commits signed off. I am not treating the external Vercel fork-authorization failure as a code failure. The two current-head review threads above remain unresolved, so this head is not approvable or mergeable under the review policy. After the provenance/classification fixes land on a new exact head and the applicable threads are reconciled, it needs exact-head re-review.

cyjin-yl (changes requested):
Reviewing exact head `c25642894b513f98af74cf08e786c6af7432d9e3`.

The prior floating-revision blocker is addressed on this head: the compared Checkbox revision is now pinned, and the old contradictory “remaining gap” wording is gone. I’m still requesting changes because the current-head spec/evidence is not yet internally consistent:

1. `P-SHADCN-CHECKBOX-INDICATOR` now classifies `flex items-center justify-center` as an intentional adoption of the later official recipe, but the ownership is different. In the pinned later recipe, those centering utilities are on `.cn-checkbox` (the Root), while `.cn-checkbox-indicator` only sizes the glyph. In the current Proto implementation, `packages/prototypes/shadcn/src/checkbox/root.proto.ts` has no flex-centering tokens and `packages/prototypes/shadcn/src/checkbox/indicator.proto.ts` owns `flex`, `items-center`, and `justify-center`. The current 14px geometry can make the rendered result look equivalent, but that does not establish adoption of the same centering mechanism/owner; composition or dimension changes can expose the difference. Please either keep this as an explicit ownership/composition deviation, or move the centering responsibility to Root and add evidence that covers the intended ownership. Do not record it as the same adopted mechanism while the anatomy owner differs.

2. `packages/spec/graph/test/shadcn-checkbox-baseline.test.ts` now protects the later comparison SHA `63c1308d112b6b1205d86244a156cca1abef5087` and rejects the floating `current upstream main` wording, but it never asserts the historical compatibility baseline `f31ed81983653919dd4fe77aee4b4859f610f1dc`. A later edit can therefore advance the historical baseline while preserving every current assertion. Please bind the test to the historical revision/source as well, on both applicable entities if that is the intended invariant.

3. The current spec rationale for keeping `rounded-[4px]` still refers to an issue ruling without making that authority reconstructible from the entity itself. The PR context points to #379, but the catalog evidence should identify the exact ruling/record/decision (or rewrite the text so it does not present an anonymous ruling as authority).

Exact-head repository CI is green (`CI` run 33518275791). I am not treating external preview authorization as a code failure. However, the three non-outdated current-head review threads above are unresolved, and the old `CHANGES_REQUESTED` cannot be superseded by an approval until these findings are reconciled on a new exact head.

guangliang2019 (dismissed):
Reviewed exact head `c25642894b513f98af74cf08e786c6af7432d9e3`.

- **[P2] PR583-CENTERING-OWNER-001** (spec/prototypes/P-SHADCN-CHECKBOX-INDICATOR.yaml:81) The entity classifies flex centering as adoption of the later recipe, but that recipe assigns flex centering to Root (.cn-checkbox), while Proto UI assigns it to Indicator. Expected: Compatibility classification must preserve the anatomy owner of the mechanism or explicitly classify the ownership/composition difference as a deviation. Fix: Record the owner difference as an explicit deviation, or move centering to Root and add evidence for the intended ownership.
- **[P2] PR583-BASELINE-GUARD-002** (packages/spec/graph/test/shadcn-checkbox-baseline.test.ts:38) The guard pins the later comparison revision but does not assert historical compatibility baseline f31ed81983653919dd4fe77aee4b4859f610f1dc. Expected: Executable evidence must bind both revisions on which the catalog comparison depends. Fix: Assert the historical baseline revision/source on each applicable entity.

cyjin-yl (changes requested):
Reviewing exact head `bc6607c35547b8bb745e3049e267796d3a6f2f4e`.

The previous owner-classification and historical-baseline guard findings are materially addressed on this head: the Root/Indicator ownership difference is now explicit, both pinned revisions are bound in the source ledger and focused test, and trusted repository `CI` run `33692741664` is green. I am not treating the external Vercel fork-authorization status as a code failure.

I’m still requesting changes for two current-head spec/evidence inconsistencies:

1. `spec/prototypes/P-SHADCN-CHECKBOX.yaml` now explicitly adds the Root-versus-Indicator centering ownership difference as an intentional deviation, but the same bilingual compatibility inventory still says only “Five intentional deviations are decided” before enumerating the pre-existing deviation set. That leaves the authoritative inventory internally contradictory. Please either update the count consistently in both locales or, preferably, restructure the inventory so every intentional deviation is named without a fragile numeric total. Add/adjust the focused regression so introducing another classified deviation cannot leave the inventory summary inconsistent.

2. The earlier radius-provenance blocker is still present verbatim: the context says the later `rounded-[6px]` recipe is not adopted because `rounded-[4px]` is retained “per the issue ruling”, but this entity still does not identify a reconstructible issue comment, decision record, or catalog relation that supplies that authority. The upstream files prove what the two recipes contain; they do not prove a maintainer ruling. Please either bind the exact ruling/decision as durable evidence or remove the anonymous-authority wording and state the current 4px choice only from evidence the entity itself owns.

The new exact-head Codex thread about the deviation total remains non-outdated and unresolved, and the radius finding remains applicable despite its older thread being outdated by line movement. This head is therefore not approvable yet. After both findings are fixed on a new exact head and applicable threads are reconciled, it needs exact-head re-review.

cyjin-yl (changes requested):
Re-reviewed exact head `8c19296050f0933c1ccbbf014c66bbdf86dcfbba`.

The two blockers from my previous review are fixed on this head: the inventory no longer hard-codes a stale deviation count, and the radius decision now has a concrete reconstructible ruling reference. Trusted repository CI is also green on this exact head.

One blocking evidence-registration gap remains. This revision adds `packages/spec/graph/test/shadcn-checkbox-baseline.test.ts` as the executable guard that pins the historical baseline, compared revision, adopted recipe, radius ruling, and named deviations. However, `spec/tests/T-SHADCN-CHECKBOX-0001.yaml` on this same head still only maps its existing prototype/browser implementations; it has no case/implementation entry for the new graph guard. Since `P-SHADCN-CHECKBOX` reaches the evidence through its `verifies` relation to this T entity, the new required guard is not discoverable through the catalog evidence graph. Marking the earlier inline thread resolved does not change that repository state.

Please register the graph guard in `T-SHADCN-CHECKBOX-0001` using the repository's normal test-entity conventions: add a dedicated (or otherwise correctly mapped) case and an implementation entry with the exact path, required status, and accurate `consumesCases`/`exercises` linkage. The acceptance criterion is that traversal from the Checkbox P entity through its T evidence reaches this executable guard, with catalog/graph validation proving the mapping rather than relying on an uncataloged test file.

cyjin-yl (changes requested):
Re-reviewed the new exact head `b898e6473ba6d3f5d350c0d33c0c994829114145` after the post-review push.

The blocker I raised on the previous head is fixed: `T-SHADCN-CHECKBOX-0001` now has a dedicated compatibility-record case and a required implementation entry pointing to `packages/spec/graph/test/shadcn-checkbox-baseline.test.ts`, so the guard is catalog-reachable. Trusted repository CI is green on this exact head.

One current-head traceability gap remains in that required guard. The Root ledger is protected against loss of all three upstream anchors (`PINNED_BASELINE`, `COMPARED_REVISION`, and `ADOPTED_RECIPE`), but the Indicator ledger only asserts the first two. The Indicator's new centering classification specifically depends on the adopted `1773ecfe...` recipe; if that source is removed or changed from `P-SHADCN-CHECKBOX-INDICATOR.sources`, the required guard still passes while the claim has lost the evidence that justifies it.

Please add an `ADOPTED_RECIPE` assertion for `indicatorSources` (and keep the existing historical/comparison assertions). The verifiable acceptance criterion is that removing any of the three upstream evidence anchors needed by the Indicator compatibility record makes the required workspace check fail. There is also a current non-outdated unresolved inline thread for this exact finding; reconcile it once the test is updated.

cyjin-yl (approved):
Re-reviewed exact head `6e9562f772c041e97f90962c524af0141d738a97`.

The remaining compatibility-ledger blocker is fixed on this head: the required `shadcn-checkbox-baseline.test.ts` guard now asserts `ADOPTED_RECIPE` on the Indicator source ledger as well as the pinned historical baseline and exact compared revision. The earlier deviation-count and radius-provenance findings are also preserved by executable assertions, and `T-SHADCN-CHECKBOX-0001` now registers this graph test as a required workspace check.

All current review threads are resolved. Trusted repository CI run `33699021617` is green (`test`, `type-check`, `lockstep-check`, `release-scan` and applicable lanes), and DCO reports all commits signed off. I am not treating the external Vercel fork-authorization status as a repository CI failure.

I found no remaining spec/evidence or implementation blocker on this exact head.

Issue #584, opened 2026-09-01
[discussion] Feasibility of a compiler mode replacing the runtime adapter


HyacinthHaru:
## What this is

A **feasibility discussion**, not a proposal to build anything. No decision is being asked for and no work is claimed.

The question: Proto UI runs `Prototype protocol -> Adapter -> Host component instance` at runtime. How far could a **compiler mode** go instead — compiling a prototype ahead of time into host-native source, the way Svelte compiles or Mitosis transpiles, so a consumer ships generated React/Vue/WC code with little or no runtime?

I researched it against the codebase rather than in the abstract. Findings that argue against are included at equal weight; the open questions at the end are things I could not answer from the code and did not guess at.

## Verdict

**Partially feasible under narrow scoping.** The declarative *projection* channels can move to compile time. The *coordination* layer cannot. And the authoring language is not a compilable subset.

The useful framing is not "replace the adapter" but **"widen the compile-time seam that already exists"** — today 1 of 9 `def` channels is partially lowered.

## The most informative evidence is what is already compiled

Proto UI already has one ahead-of-time channel, and studying it beats speculating.

`packages/cli/src/services/prototype-style-tokens.ts` (778 lines, `@ts-nocheck`, `ts.createSourceFile` — **syntax only, no type checker**) lowers `def.rule({when, intent})` into variant tokens. `packages/modules/rule-expose-state-web/src/create.ts` does the same lowering at runtime and then *removes* the lowered rules from the runtime set.

Three things this proves:

**1. The static half cannot decide — it over-approximates.** The runtime's `extractConditions` handles `eq` and `all` and returns `null` for everything else, so a rule using `w.any(...)` is never lowered. The static extractor *does* descend into `any` and joins with `:`, which is Tailwind's **conjunction**. The result is checked into the repo:

```
packages/cli/src/generated/shadcn-style-tokens.ts:59
  'data-[active]:data-[focus-visible]:data-[focused]:data-[hovered]:bg-accent'
```

That is a four-way **AND** generated from the four-way **OR** at `packages/prototypes/shadcn/src/select/item.proto.ts:44-50`. I verified it is dead CSS — the runtime never applies it. The stylesheet is still correct only because the extractor separately picks up the bare `bg-accent` from the intent arrow and emits that too.

So after building a 778-line AST extractor, the project still cannot statically answer "will the runtime lower this rule?", and copes by **emitting both answers**. A compiler that must *replace* the runtime does not get that escape hatch.

[...]

PR #585, opened 2026-09-01
fix(cli): give Select Trigger placeholder a lowered variant

HyacinthHaru:
Found while researching #584. It is a live styling defect, so it stands on its own; the discussion is context, not a prerequisite.

## The defect

`packages/prototypes/base/src/select/trigger.proto.ts:22-23` exposes `placeholder` as a bool state. Both Select Trigger projections key a rule on it:

```ts
def.rule({
  when: (w) => w.state(placeholder).eq(true),
  intent: (i) => i.feedback.style.use(tw('text-muted-foreground')),
});
```

The runtime lowers that correctly and writes the class onto the host. Measured on a mounted shadcn Select Trigger:

```
data-placeholder
data-pui-style="… data-[placeholder]:text-muted-foreground data-[hovered]:bg-input/50 …"
```

But the CLI extractor produced **no** variant token for it, because `resolveKnownAsHookStateHandles` routes `asSelectTrigger` through the shared five-state command map, which has no `placeholder`. So `data-[placeholder]` appeared **zero** times in either generated closure, and no CSS rule matched the class that was actually applied.

Net effect: **the Select Trigger placeholder was never muted, in shadcn and in Brutalist.** The bare `text-muted-foreground` is in the closure, but `data-pui-style~=` is whole-word, so it cannot match the variant token either.

The fix is one map entry — `asSelectTrigger` gets its own branch with `['placeholder', 'data-[placeholder]']` — plus the regenerated closures, which gain exactly one token each.

## Why it went unnoticed, and the gate

This failure mode is silent by construction. A missing entry does not crash and does not fail `check:styles:preset`, because that check only compares the extractor against its own prior output — if the extractor never emitted the token, the manifest never had it, and the two agree forever.

I have now hit the same shape three times:

| where | consequence |
| --- | --- |
| #472 | Checkbox rules produced no variants; the dark tint reached the closure as an unguarded `dark:bg-input/30` covering the filled box |
| #482 | `ring-2 ring-inset` reached the closure unconditional, which would have painted the focus ring at rest |
| here | the applied class had no CSS at all |

The first two I caught by reading the generated diff. That is not a control.

`packages/cli/test/lowered-hook-coverage.test.ts` makes it one. It walks every shipped `*.proto.ts`, traces each `w.state(x)` in a rule condition back to the `asHook().stateHandles` binding it came from — including the two-step `const s = asHook().stateHandles; const { x } = s;` shape the prototypes actually use — and asserts the extractor can resolve that exact hook/state pair.

It asks the resolver itself through a new `loweredHookStates()` probe rather than keeping its own copy of the hook list, so the gate cannot drift from the table it guards. It also asserts the scan finds more than twenty usages, so a refactor cannot quietly empty it into a vacuous pass.

Run against `main` it reports:

```
packages/prototypes/shadcn/src/select/trigger.proto.ts: asSelectTrigger().placeholder
packages/prototypes/brutalist/src/select/trigger.proto.ts: asSelectTrigger().placeholder
```

## Scope

Deliberately narrow. This does not replace the hand-written table with a generated one — that is the larger question in #584, and doing it properly needs a type checker the CLI does not currently ship. The gate closes the failure mode that has actually bitten, and leaves coverage at 21 of 47 hooks where it is.

I have not audited whether other *uncovered* hooks would misbehave if a future rule keyed on them; the gate will now say so at the moment it happens, which is the point.

## Verification

- Defect reproduced on a mounted component before the fix, and the rendered CSS confirmed present after (`:where([data-pui-style~="data-[placeholder]:text-muted-foreground"])[data-placeholder]`).
- Gate verified red before the fix with the two entries above, green after.
- `vitest run packages`: 1481 passed, 34 todo, 3 skipped.
- `check:styles:preset`, `check:styles:variant-order`, `check:prototype-catalog`, `check:public-docs`, `check:agent-doc`, `check:types` all green.


cyjin-yl (changes requested):
Reviewing exact head `d4d5c7ef832a7c5c171d10718a598f42bebd37ca`.

I’m requesting changes for two blocking items.

1. The Select Trigger placeholder mapping itself is narrow and the regenerated `data-[placeholder]:text-muted-foreground` tokens are consistent with that mapping, but the new coverage gate does not actually cover the binding shapes it claims to guard. In `packages/cli/test/lowered-hook-coverage.test.ts`, `hookNameOf()` only recognizes a direct `asHook().stateHandles` expression, and `visitBindings()` never records a hook-result binding such as `const hook = asHook()`. The exact reviewed tree already contains this shape in `packages/prototypes/brutalist/src/textarea/root.proto.ts`: `const textarea = asTextareaRoot(); const state = textarea.stateHandles; ... w.state(state.focusVisible)`. That usage is omitted from `usages`, so removing or drifting the corresponding resolver entry can still leave this gate green. Please track hook-result bindings through the subsequent `.stateHandles` access (with lexical-scope-safe binding), and add a focused regression/fixture that demonstrates the gate turns red when a resolver pair used through this two-step shape is removed.

2. The trusted repository `CI` run for this exact head, `33517661697`, is failing. The `test` job fails in `Public documentation gate and repository tests`; `type-check`, `lockstep-check`, and `release-scan` succeeded. Required repository CI must be green on the corrected exact head before approval. I am not counting the external Vercel fork-authorization status as a code failure.

The commit itself carries a valid `Signed-off-by` trailer. After the scanner blind spot is fixed and repository CI is green, please request exact-head re-review; this head is not approvable or mergeable under the current policy.

cyjin-yl (changes requested):
Reviewing exact head `a40e7e304124e0dbdc6fdbb97021a154b511101a`.

The previous two-step hook-result blind spot is materially improved, and trusted repository CI run `33584690316` plus DCO are green. I am not treating the external Vercel authorization status as a code failure.

I’m still requesting changes for two current-head coverage-gate problems:

1. In `packages/cli/src/services/lowered-hook-coverage.ts`, `readStateLeaf()` returns `null` when a lowerable state read uses a binding shape the scanner does not understand, and `visitRules()` then silently drops that leaf via `if (leaf.usage)`. That keeps the gate fail-open: a new syntax such as `const checked = asCheckboxRoot().stateHandles.checked` (or an aliased `asHook` binding) can be missed by both the static extractor and this coverage scan while the runtime rule remains semantically eligible for lowering. Please make “lowerable but unresolved” a first-class scan result, fail the repository-wide gate on any such leaf, and add a regression proving an unsupported form turns the gate red instead of disappearing from `found`.

2. The binding tables are still whole-file `Map<string, ...>` structures keyed only by identifier text and populated by repeated global passes. That is not lexical-scope-safe: a shadowed hook/handle-bag/handle identifier in a nested or sibling scope can overwrite another binding and misattribute a rule to the wrong hook. The previous review explicitly required lexical-scope-safe tracing. Please use scope-aware binding identity/environments and add a nested/sibling shadowing regression that would fail with the current global maps.

Once both are fixed, the applicable current threads are reconciled, and the corrected exact head retains green trusted CI/DCO, please request exact-head re-review.

guangliang2019 (dismissed):
Reviewed exact head `a40e7e304124e0dbdc6fdbb97021a154b511101a`.

- **[P2] PR585-SCOPE-BINDING-003** (packages/cli/src/services/lowered-hook-coverage.ts:33) The scanner keys hook results, handle bags, and handles by identifier text across the whole source file. A current-head probe with outer const state = asCheckboxRoot().stateHandles and an inner block shadowing const state = asSelectTrigger().stateHandles reports the later outer state.checked usage as asSelectTrigger().checked. Expected: Bindings must be resolved by lexical symbol/scope so a nested same-name declaration cannot change the hook identity of an outer rule usage. Fix: Track declarations and uses through TypeScript symbols or an explicit lexical-scope stack, and add a nested-shadowing regression that proves both inner and outer usages retain their own hook identities.

cyjin-yl (changes requested):
Reviewing exact head `a99091b5122cdde30a1b3640b97a3777812e753a`.

The two blockers from my previous review are materially improved: the scanner now fails closed on untraceable lowerable reads and uses lexical scope rather than whole-file identifier maps, with nested/sibling regressions. Trusted repository `CI` run `33692600265` is green (`test`, `type-check`, `lockstep-check`, `release-scan`, package/release lanes), and DCO reports all commits signed off. I am not treating the external Vercel authorization status as a code failure.

I’m still requesting changes because the new gate is not yet aligned with the production extractor in three places:

1. `packages/cli/src/services/lowered-hook-coverage.ts` now resolves `const checked = asCheckboxRoot().stateHandles.checked` as a covered hook-state usage, but `packages/cli/src/services/prototype-style-tokens.ts` does not resolve that terminal semantic-map leaf when `registerDeclaration()` calls `resolveBinding()` / `resolveExpression()`. The extractor only special-cases a property access named `stateHandles`; a terminal `.checked` binding becomes empty, so `w.state(checked)` can still emit no lowered token while the coverage gate is green. Either teach the production extractor this binding form and add an end-to-end token regression, or keep it first-class unresolved in the gate until the extractor supports it.

2. The coverage scanner treats any `.eq(...)` RHS other than the `false` keyword as a positive lowerable leaf. The production `resolveStateEqVariant()` only lowers boolean keywords and string literals. Numeric or bound-identifier comparisons such as `w.state(step).eq(1)` / `w.state(checked).eq(ENABLED)` can therefore pass the gate while producing no static variant. Align the scanner with the extractor’s accepted RHS syntax, or extend the extractor and prove the additional forms end to end; unsupported lowerable RHS forms must fail closed rather than count as covered.

3. The repository-wide test scans only `packages/prototypes/*/src/**/*.proto.ts`, while the production source collector accepts the broader JS/TS module family (`.ts`, `.tsx`, `.mts`, `.cts`, `.js`, `.jsx`, `.mjs`, `.cjs`). A lowerable rule in another production-consumed source extension is invisible to both `usages` and `unresolved`, so the aggregate gate can remain green. Make the coverage scan operate over the same production source set/extension policy rather than maintaining a narrower glob, with at least one non-`.ts` fixture or integration regression.

All three corresponding exact-head review threads are non-outdated and unresolved. Once the coverage analyzer and production extractor share these boundaries, the corrected head retains green trusted CI/DCO, and applicable threads are reconciled, please request exact-head re-review.

cyjin-yl (commented):
Re-reviewed exact head `a96191a15e36da0f1094717c58d05a03ecda2cc5`.

The three code blockers from my previous review are addressed on this head: production expression resolution now follows chained state-handle property access, the coverage scanner fails closed on unsupported comparison RHS forms, and the repo-wide scan uses the production source-file collector rather than a narrower `*.proto.ts` glob. The new focused coverage tests pass as part of the repository test job.

I cannot supersede the previous change request with an approval yet because trusted repository CI on this exact head is red. CI run `33697320124` fails in `apps/www/src/content/docs/zh-cn/demo-shadcn-controls.browser.test.ts` at the React transition-property assertion: expected `color, box-shadow`, received `none`. This is actual repository CI, not Vercel/preview authorization. The failure is outside the files changed by this PR, so I do not have enough evidence to claim a new implementation defect here; however, it also reproduced on the fresh run the author explicitly pushed for that purpose.

No additional code change is requested from the inspected extractor increment at this point. The remaining acceptance criterion is a successful trusted repository CI run on this exact head. If the browser failure reproduces again, it needs to be diagnosed/fixed (or otherwise demonstrated to be an independent repository failure) before this head can be approved or merged.

cyjin-yl (changes requested):
Re-reviewed exact head `87928a63b8f4ec78bc91d0dd0f143cd50e5fb609`.

The earlier extractor/scanner alignment findings are materially fixed, and trusted repository CI run `33701368484` is green. One fail-open gap remains in the new gate:

`packages/cli/src/services/lowered-hook-coverage.ts` decides whether an intent can yield extractable tokens through `resolvableTokenArgument()`, and currently treats every Identifier, PropertyAccessExpression, and ElementAccessExpression as resolvable without consulting the production extractor's lexical/import bindings. Production `resolveExpression()` does not make that guarantee: an identifier only yields tokens if `lookup()` has a value, and `applyImportBindings()` only resolves named imports from relative modules. A valid prototype can therefore use, for example, a non-relative imported token constant (or a local constant initialized from an opaque helper) in `tw(TOKENS)`. The runtime receives the real token string, but the static extractor resolves the identifier to an empty value; meanwhile this coverage scanner returns `true`, records the hook/state as covered, and the gate stays green while the prefixed variant CSS is missing.

Please make intent-token coverage ask the same production resolution path/scoped bindings that `collectTwTokens`/`resolveExpression` use, or fail closed for identifier/property/element forms unless the scanner can prove the value is statically resolvable. Add a focused regression where `tw(...)` receives a type-valid but production-unresolved bound/imported value and prove the repository gate reports it unresolved (or extend production extraction and prove the variant end to end).

This is the same silent-failure class the gate is intended to close, so I consider it blocking rather than a test-style nit. I am not treating external Vercel authorization as a code failure. After this alignment is fixed on a new exact head, trusted CI/DCO remain green, and the applicable review state is reconciled, it can be re-reviewed.

cyjin-yl (changes requested):
Re-reviewing exact head `8cd6350d31cb80ae12da3e0f0f4845adc1539047`.

The blocker from my previous `87928a63...` review is materially addressed: intent coverage now resolves local token bindings, follows relative imported initializers, rejects opaque/non-relative token sources, and validates every `tw(...)` argument. The earlier extractor/scanner findings are also represented by resolved threads.

I am still requesting changes because the current head has four new analyzer-alignment gaps, all in the same correctness boundary this gate is intended to enforce:

1. `resolvableTokenArgument()` accepts any `.join(...)` once the receiver is resolvable, but production `resolveJoinCall()` only honors a literal/no-substitution-template separator and otherwise defaults to `,`. For `const SEP = ' '; tw(['bg-a', 'text-b'].join(SEP))`, the gate therefore marks the intent extractable while production emits the comma-joined static token and runtime evaluates the space separator. Validate the separator with the same production rule (or share the resolver) and add a bound-separator regression.

2. The relative-import check reconstructs an imported module scope from local variable declarations only. Production `loadModuleBindings()` first applies that module's own relative imports recursively. A chain such as `style.ts: import { BASE } from './base'; export const TOKENS = BASE` is production-extractable but this gate reports `BASE` unresolved and blocks valid code. Build imported-module scopes with equivalent recursive import bindings, with a transitive-relative-import regression.

3. Both the scan and production AST walk recognize only property-access `def.rule(...)`. A typed call through literal element access, `def['rule']({ ... })`, invokes the same runtime API but contributes neither `usages` nor `unresolved`; the extractor emits no variant while the runtime can lower the rule and remove its fallback. Either support literal element access in both paths or make the coverage gate fail closed on it.

4. Function scopes are created, but parameter bindings are not installed as lexical shadows. An inner function parameter reusing an outer state-handle identifier can therefore fall through to the outer binding in both analyzers and approve/emit the wrong state variant. Parameters must shadow outer bindings even when their state-handle origin cannot be statically recovered; unresolved origins should fail closed. Add a parameter-shadowing regression.

Separately, trusted repository CI run `33704524557` is red on this exact head: the `test` job fails in `Public documentation gate and repository tests`, while `type-check`, `lockstep-check`, and `release-scan` pass. The repeated React transition assertion appears to be the known repository flake discussed around #595, so I am not attributing that failure to this diff without evidence, but required repository CI still has to be green before approval. The four non-outdated review threads above are also currently unresolved.

Once the analyzer boundary is aligned on a new exact head, the threads are reconciled, and trusted CI/DCO are green, request exact-head re-review.

cyjin-yl (approved):
Approving exact head `3187ef78f045d7cab0ddf6902268075c2e8afd00` after re-reviewing the latest increment and reconciling the live findings.

The two current-head gaps from `183374d8` are fixed in this revision: helper-produced condition branches are now rejected rather than partially accepted, and template substitutions must be single-valued in the same cases where the production extractor can actually reproduce the token. The focused regressions cover both paths. The earlier scanner/extractor alignment findings across binding forms, source extensions, scopes, condition syntax, and intent extraction remain incorporated in the current tree, and all applicable review threads are resolved.

The exposed prototype-owned-state mismatch discovered during this review is real but predates and exceeds this PR's Select Trigger / hook-table coverage slice. I independently verified it and have now given #600 a separate bounded implementation ruling that requires the static extractor and coverage gate to learn exposed-local state semantics together. Keeping that product fix out of this already deep test/gate PR is the cleaner scope boundary and does not invalidate the placeholder fix or the hook-state gate for the class it is designed to cover.

Trusted repository CI is green on this exact head (`test`, `type-check`, `lockstep-check`, `release-scan`, package/release smoke lanes), and DCO reports all commits signed off. No current blocking thread remains.

PR #597, opened 2026-09-02
test(docs): cover Brutalist Button live theme in three runtimes

HyacinthHaru:
Closes part of #539 — the Button slice only. Dialog and Hover Card, the other two the issue names as highest impact, stay for follow-up PRs.

## What was actually missing

#539 says Button has no browser evidence for a `P-BRUTALIST-BUTTON-THEME-REPAINT` criterion. That criterion id does not exist in the repository; the real one is `P-BRUTALIST-BUTTON-LIVE-THEME`, and it already has a test — `packages/prototypes/brutalist/test/button-live-theme.test.ts`, bound by `T-BRUTALIST-BUTTON-0001-CASE-8`.

The gap is narrower and is written down in that test's own file header: happy-dom implements neither CSS `@layer` nor enough `:root` cascade, so the module test regex-strips the `@layer proto-ui { }` wrapper off the token CSS and applies the theme as inline custom properties on a wrapper element. Inline styles are the highest-precedence non-`!important` source, so they win by construction; on a real page `:root.dark` competes with `:root` inside the same layer, and the token CSS sits in `@layer proto-ui`, below unlayered CSS. It also mounts Web Components only.

So the mechanism is proven, but not in the cascade it ships in, and not in React or Vue.

## The case

`T-BRUTALIST-BUTTON-0001-CASE-LIVE-THEME-BROWSER`, in the existing `demo-brutalist-controls.browser.test.ts`, over Web Components / React / Vue in Light and Dark:

- every fill equals the `:root` value of the variable it names, both painted through the same 1x1 canvas;
- `surface` moves between the two schemes, and returns exactly when the scheme returns;
- `solid/main` is byte-identical across schemes, enabled and disabled;
- the disabled `surface` control carries the same pair as the enabled one at `opacity: 0.5`;
- `data-hovered` and `data-pressed` are absent throughout and the pointer is parked off the demo, which is the "without pointer interaction" half of the criterion.

Comparing each fill against the variable it names, rather than against a pinned colour, is deliberate: the case then states *which variable the surface follows* instead of restating the palette, and does not fail every time a theme value is retuned.

## Two things I changed that are worth arguing about

**The demo gained a disabled `surface` Button.** The disabled example it already had is the default `solid/main`, whose accent pair is theme-invariant, so it cannot show a repaint at all — the criterion's disabled sentence had nowhere to land. The existing disabled example is kept. This is the same move as widening the Scroll Area demo in #482 so one surface could scroll both ways. If you would rather freeze the demo, I will drop the disabled assertions and the criterion keeps that sentence unproven in a browser.

**`destructive` is not asserted to change.** `--pui-destructive` is `#fecdd3` in both schemes, with `#000000` foreground. `P-BRUTALIST-BUTTON-LIGHT-DARK` reads "Surface and destructive fills resolve through CSS variables so their concrete colors follow the active host theme" — the resolution happens, but the value does not move, so only the mechanism half is true for `destructive`. I have asserted the mechanism and left the palette alone; retuning it is a design decision, not a test fix.

## Evidence

Mutation-tested rather than merely passing. Replacing `surface`'s `bg-secondary-background` with `bg-main` — a token that is in the closure and paints normally — fails at `wc/light/surface/background: expected '254,240,138,255' to be '255,255,255,255'`. The first attempt used `bg-white`, which failed at `'0,0,0,0'`: it is not in the Brutalist token closure, so it generates no CSS at all, which is the silent class #585 is about.

- Brutalist browser suite: 9/9, including the 8 pre-existing cases.
- `check:prototype-catalog`, `check:types`, `check:agent-doc`, `check:public-docs`: green.
- Non-browser suite: 1515 passed. The 29 failures on my machine are all one `localStorage is undefined` root cause in `apps/www` component tests, reproduce identically on a clean `origin/main`, and come from Node 26 shadowing happy-dom's `localStorage` unless `--localstorage-file` is passed. Nothing to do with this branch.

`scripts/test/run-runtime-tests.mjs` gained the Button route so the shared dev server warms it, matching the comment already there about cold Astro routes.


cyjin-yl (changes requested):
Reviewed exact head `c77a874d23107ccc2124f2218c131d1490d657fc`. Trusted repository CI is green, but the new browser evidence has two blocking coverage problems.

1. The live-theme test drives the transition through `applyColorScheme()`, which calls Playwright `emulateMedia({ colorScheme })`. That does not prove the contract this test claims to verify. `P-BRUTALIST-BUTTON-LIVE-THEME` requires an already-mounted Button to repaint when the active host theme changes, and explicitly treats `colorScheme`/`prefers-color-scheme` as outside that criterion. A renderer accidentally coupled to media preference could therefore pass this test while failing the governed host-theme path. Drive the actual host/document theme surface that owns the CSS variables (the project's `data-theme`/host-theme mechanism), independently of emulated media, and show the same mounted Button surfaces repaint across light→dark→light, including the disabled case.

2. The test's `RUNTIMES` matrix is `wc`, `react`, `vue`, while the public runtime registry is `wc`, `react`, `vue`, `vue2`. The new T case and assertions describe this as evidence across every runtime, so Vue 2 is currently untested. Include `vue2` (preferably derive the browser evidence matrix from the canonical adapter IDs if practical) and ensure the demo refs/assertions exercise it as well.

Acceptance is a host-theme-driven, no-remount light/dark recovery assertion across all four official web runtimes, with the T implementation mapping continuing to point at that executable evidence.

cyjin-yl (approved):
Re-reviewed exact head `e0bffaa68e0951e4f853419eb107c04490a462d2` after the fixes.

Both blockers from the previous review are resolved. The browser case now pins media preference to Light and drives the actual documentation host-theme API (`StarlightTheme.set` / `data-theme`) without adding a `.dark` class, then proves light→dark→light repaint on the same mounted controls. It also uses the shared four-runtime harness, including Vue 2. The canary phase independently moves the surface/destructive variable pairs while leaving the fixed solid accent pair alone, so hard-coded palette values cannot satisfy the required variable-backed evidence.

The required `T-BRUTALIST-BUTTON-0001` implementation points to this executable browser case, all review threads are resolved, trusted repository CI run `33701201199` is green, and DCO reports all commits signed off. The merge of independent PR #583 changed `main` after this head's CI run but does not overlap this PR's five-file diff or introduce a prerequisite; GitHub still reports the PR mergeable.

I found no remaining blocking runtime/spec/evidence issue on this exact head.

Issue #598, opened 2026-09-02
audit(spec): two entity source citations point at files that do not exist, and no check fails on it


HyacinthHaru:
Two `spec/**` entities cite in-repo evidence that is not in the repository. Both citations are load-clean, so `check:prototype-catalog`, the workspace loader, and every other repository check pass on `main` with them in place.

Baseline `02666e14`.

## Survey

Walking `sources[].path` and `implementations[].path` on every entity and testing existence:

| | count |
| --- | ---: |
| entities | 549 |
| source citations | 1365 (1255 in-repo, 110 URLs) |
| **in-repo source citations that do not resolve** | **2** |
| implementation paths | 475 (all in-repo) |
| implementation paths that do not resolve, `status: planned` | 6 |
| implementation paths that do not resolve, any other status | 0 |
| loader issues (`workspace.issues`) | 0 |

The 6 planned ones are fine and must stay allowed — a `planned` implementation is a declared intent, not a claim that a file exists. The 2 source citations are the finding.

## 1. `C-DELAY-0001`

```yaml
sources:
  - path: internal/records/2026-07-06-presence-visibility-governance.zh-CN.md
    sections:
      - 近期工作路线
      - 保留断口
```

That record has never existed on any branch — `git log --all --diff-filter=A` finds no commit adding it. Neither section heading exists in any file under `internal/records/`, so the citation cannot be repaired by guessing a rename either. Five other `2026-07-06-*` records do exist; none carries those headings.

Introduced by `f2980543` (`feat: add core delay primitive`, 2026-07-07). Most likely the record was written alongside the primitive and never committed.

## 2. `D-AS-HOOK-DEFINE-HOOK-0001`

```yaml
sources:
  - path: packages/prototypes/base/src/behaviors/use-focus-roving.ts
```

This one is a stale reference rather than a missing artifact: the file was removed by `0d8c1383` (`refactor: catalog select prototypes`, 2026-07-18). The repository already holds both halves of the successor evidence — `packages/hooks/src/as-focus-roving.ts` and `internal/records/2026-07-06-use-focus-roving-deprecation.zh-CN.md` — so the decision entity is pointing at the pre-refactor spelling of something that still exists under a new identity.

## Why this matters beyond two lines

[...]

PR #599, opened 2026-09-03
spec(catalog): require cited evidence to exist and cases to be consumed

HyacinthHaru:
Implements the two bounded authorizations on #598: path integrity, and case-consumption graph integrity. Both deferred questions — `sections` anchor validation and a planned-path expiry policy — stay out.

## The two broken citations

**`C-DELAY-0001`** cited `internal/records/2026-07-06-presence-visibility-governance.zh-CN.md`, which has never existed on any branch and whose two named sections appear in no record, so there was nothing to recover. Per the ruling I did not reconstruct it. The entity is re-sourced to evidence that exists and is auditable: `packages/core/src/delay.ts`, `packages/runtime/test/contract/delay.v0.contract.test.ts`, and `packages/adapters/base/test/adapter-host-delay.test.ts`. No `sections` on those, since none has a stable addressable section this record needs.

**`D-AS-HOOK-DEFINE-HOOK-0001`** cited `packages/prototypes/base/src/behaviors/use-focus-roving.ts`, removed by `0d8c1383`. Replaced with `packages/hooks/src/as-focus-roving.ts` — `sections: [asFocusRoving]`, verified against the exported symbol at line 20 — plus the deprecation record that carries the migration history. No pretend-current historical path.

## The guard

`packages/spec/graph/test/catalog-evidence-integrity.test.ts`, three invariants over the loaded workspace:

- every non-URL `sources[].path` resolves;
- every `implementations[].path` resolves unless that implementation is `status: planned`;
- every declared case is consumed by at least one implementation.

Diagnostics name the entity and the offending path or case. The focused fixture drives all four cases the ruling asked for — missing local source fails, URL source is allowed, missing non-planned implementation fails, missing planned implementation is allowed — plus an unconsumed case, in one hand-built entity with no repository scan.

## The six orphan cases

Reconciled individually, not bulk-labelled. I checked each expectation against the candidate tests rather than trusting a keyword match, and all six turned out to need `planned` implementations:

| case | why planned |
| --- | --- |
| `T-AS-HOOK-PRIVILEGED-0001-CASE-ROLLBACK-BOUNDARY` | `rollback` appears in none of the four privileged contract tests |
| `T-FOCUS-0001-CASE-FOCUS-VISIBLE-NATIVE-EQUALITY` | the focus contract sets `focusVisible` in facts snapshots without comparing it to the host native result; no adapter test asserts the modality-fallback half |
| `T-LIFECYCLE-0002-CASE-RENDER-RUNTIME-INPUT` | the nearest evidence proves a mounted set does not re-render until an explicit update — the absence of an automatic render, not a read during render that provably does not re-enter |
| `T-LIFECYCLE-0002-CASE-COALESCED-UPDATE` | per your ruling: executable only where coalescing exists, and this entity records that it does not |
| `T-LIFECYCLE-0003-CASE-UPDATE-REVISION` | `revision` appears zero times in the session contract, which asserts a mount epoch |
| `T-RULE-WHEN-0002-CASE-META-DEFERRED` | joined the `planned` implementation this entity already carried; the decision record states the boundary and nothing asserts it |

Each planned implementation carries a note saying what was checked and why the nearest current test is not it. The `focus-visible` note also records the adjacent evidence — the documentation browser suites do compare a projected `focusVisible` against `:focus-visible`, but under their own Textarea entities and only for the measurable half — so a maintainer who disagrees can see exactly what I weighed.

## Verification

Every invariant mutation-tested rather than merely passing:

- broke a source path → `C-DELAY-0001: packages/core/src/delay-gone.ts`
- flipped a `planned` implementation to `passing` → `T-LIFECYCLE-0003: runtime-lifecycle-update-revision -> …lifecycle.update-revision.v1.contract.test.ts`
- added an unconsumed case → `T-LIFECYCLE-0003: T-LIFECYCLE-0003-CASE-UNCONSUMED-PROBE`

The second of those found a hole in my own guard. My first orphan-case mutation deleted a `consumesCases` entry, which left invalid YAML — and the test **passed**, because an entity that fails to load contributes no cases and the invariant read as satisfied. The orphan test now asserts `workspace.issues` is empty first, and that same mutation fails.

`packages` 1486 passed, `check:prototype-catalog`, `check:types`, `check:agent-doc`, `check:public-docs` all green. Repository-wide scan is clean on all three invariants.


cyjin-yl (changes requested):
Reviewing exact head `24b56ad404e0aeb5d0f6b0a36be81d2691f1cdd0`.

Trusted repository CI (`33704970393`) and DCO are green, and the PR is non-draft/mergeable, but I am requesting changes for three independently verified catalog-integrity problems.

1. `packages/spec/graph/test/catalog-evidence-integrity.test.ts` now exempts both `planned` and `missing` implementations from path existence via `DECLARED_ABSENT`. That is broader than the maintainer decision in #598: the authorized rule was **every implementation path must resolve unless the implementation is explicitly `status: planned`**, with the focused negative regression “missing non-planned implementation fails.” The current fixture instead asserts that a `status: missing` implementation may point to a nonexistent file, so the new guard can stay green on exactly the case it was authorized to reject. Please make `planned` the only missing-path exemption and restore a regression proving a nonexistent `missing`/other non-planned implementation path fails.

2. `spec/tests/T-LIFECYCLE-0002.yaml` classifies `T-LIFECYCLE-0002-CASE-RENDER-RUNTIME-INPUT` as a new planned implementation, but current repository evidence already covers this boundary. `packages/runtime/test/contract/prototype.render-syntax.v0.contract.test.ts` reads props/context/anatomy during render, asserts `renderer.update` is unavailable, and completes the host commit; `T-CORE-SYNTAX-0001` already maps that exact passing runtime test to `C-LIFECYCLE-0003-D` / `C-LIFECYCLE-0003-E`. The #598 disposition explicitly required each orphan to be checked against existing assertions and only use `planned` when evidence does not exist. Reuse/map the existing passing evidence here instead of recording duplicate future work.

3. `spec/decisions/D-AS-HOOK-DEFINE-HOOK-0001.yaml` repairs the stale source path by citing `packages/hooks/src/as-focus-roving.ts` plus the deprecation record, but leaves the decision text saying official prototypes currently use `useFocusRoving`. The cited record says the implementation/export were deleted and current prototypes no longer consume that helper. Please make that mention historical (or otherwise reconcile the present-tense statement) so the decision text and its repaired evidence do not contradict each other.

There are also current non-outdated unresolved review threads on this head; they must be reconciled before approval. I am not treating green CI as sufficient when the catalog semantics themselves are inconsistent. After these findings are fixed on a new exact head, re-run the graph scan/CI and request exact-head review.

cyjin-yl (changes requested):
Re-reviewing exact head `52096f01dc8d1563e8b9fb61b1c775bb60e31f27` after the new commit.

The incremental fixes are good: the render-input, coalescing, update-revision, and focus-visible cases are now mapped to existing passing evidence rather than duplicate planned work, and `D-AS-HOOK-DEFINE-HOOK-0001` now correctly treats `useFocusRoving` as historical. All current review threads are reconciled, trusted CI `33706372773` is green, and DCO reports all commits signed off.

One blocker remains in `packages/spec/graph/test/catalog-evidence-integrity.test.ts`: `DECLARED_ABSENT` still contains both `planned` and `missing`, so a nonexistent implementation path with `status: missing` is deliberately exempted. That contradicts the accepted #598 boundary for this slice: every implementation path must resolve unless the implementation is explicitly `status: planned`, with the focused negative regression that a missing non-planned implementation fails. The current fixture even encodes the opposite by giving `status: missing` a nonexistent path and expecting no `missingImplementations` entry.

Please keep `planned` as the only absent-path exemption and make the fixture prove that a nonexistent `status: missing` (or any other non-planned status) is rejected, while a nonexistent `status: planned` remains allowed. If `missing` is intended to acquire separate no-file semantics, that is a schema/lifecycle decision outside the bounded #598 authorization and should be decided explicitly rather than silently widened here.

After that change, re-run the exact-head graph/repository checks and request re-review.

cyjin-yl (changes requested):
Re-reviewing exact head `fbf08ae693b8a26a8eb4ea4968eb36d9962b1036` after the latest fixes. The previous path-status blocker is fixed correctly: `planned` is now the only absent-path exemption, the fixture rejects `status: missing`, pathless/remote current implementations, directory targets, traversal/absolute paths, and the repository-wide evidence/case scan is green. Trusted CI and DCO are also green on this head.

Four current-head consistency/integrity blockers remain:

1. `packages/spec/graph/test/catalog-evidence-integrity.test.ts`: `isRepositoryPath()` treats every `scheme://` value as an external citation. That means a machine-local `file:///...` source bypasses the repository-path check even though it is not portable/auditable evidence. Restrict the external exemption to the portable remote schemes this catalog actually allows (at minimum HTTP(S)), and add a regression proving a `file:` citation cannot satisfy source integrity.
2. `spec/tests/T-LIFECYCLE-0002.yaml` now correctly maps `T-LIFECYCLE-0002-CASE-RENDER-RUNTIME-INPUT` to passing evidence that reads props, context, and anatomy during render. `C-LIFECYCLE-0003-Q-RENDER-RUN-WRITE-SURFACE` still says the current renderer is props-only and context reads remain an implementation gap. Reconcile that authoritative open-question context while preserving the genuinely unresolved write-surface decision; otherwise contract and test catalog disagree about the same capability.
3. The same lifecycle entity edits the historical `0.1.0` revision summary to erase the coalescing gap, while a new `0.3.0-alpha.0` revision already records that the gap was retired. Restore the old revision text and let the new revision carry the correction so the revision history remains historical rather than retroactively rewritten.
4. `D-AS-HOOK-DEFINE-HOOK-0001` now changes the authoritative statement/relationship to record the `useFocusRoving` → `asFocusRoving` correction, but its `revisions` still ends before that correction. Add a current-version revision for this decision so the new relationship is traceable in the same way the changed T entities are.

These are all bounded to the catalog-integrity work already in this PR. Please address them, resolve the corresponding live threads, and re-run exact-head repository checks before re-review.

cyjin-yl (changes requested):
Re-reviewing exact head `0bfcf774afa6ff79cb0a86eeb0bd63ca7e75d23a` after the author's four-fix commit. The four previous findings are substantively fixed: historical coalescing text is restored, `file:` no longer bypasses repository-path validation, the render-read context now matches executable evidence, and the focus-roving decision has a current revision. All existing review threads are resolved; DCO is green. The trusted repository CI run `33717073303` is still in progress at review time, so it cannot yet support integration.

One current-head catalog-consistency blocker remains in `spec/contracts/C-LIFECYCLE-0003.yaml`. The updated open-question context now correctly says render-time props/context/anatomy reads exist, and `T-CORE-SYNTAX-0001` already maps `prototype.render-syntax.v0.contract.test.ts` as passing evidence for `C-LIFECYCLE-0003-D` and `C-LIFECYCLE-0003-E`. But `C-LIFECYCLE-0003-Q-RENDER-RUN-WRITE-SURFACE.blocks` still lists `C-LIFECYCLE-0003-D` and `C-LIFECYCLE-0003-E`. That leaves the graph simultaneously saying those criteria are executable/passing and still blocked by a question whose remaining subject is the *write* surface. Generated open-question/coverage projections will therefore continue to report already-proven read/non-recursive-render criteria as unresolved.

Please reconcile the `blocks` list so it contains only criteria genuinely left open by the write-surface decision (in particular, do not keep D/E blocked now that their evidence is explicitly accepted). Because this PR changes the authoritative meaning/state of that open question, also add a current-version revision to `C-LIFECYCLE-0003` recording the correction, just as this PR now does for the affected T entities and `D-AS-HOOK-DEFINE-HOOK-0001`. Then rerun the exact-head graph/repository checks and request re-review. No other blocker found in the current diff.

cyjin-yl (approved):
Re-reviewed exact head `7e61963793417ca47c296bde97b1930de9c2a51a`. The previous `C-LIFECYCLE-0003` blocker is resolved: the render/write open question no longer blocks criteria D/E that already have accepted executable evidence, and the current revision records the reconciliation without rewriting the earlier history.

I also rechecked the rest of this bounded slice: only `planned` implementations may omit a repository path; local/absolute/traversal/`file:` and untracked evidence paths are rejected; declared cases must be consumed; the two stale citations are repaired; the orphan cases are reconciled against actual evidence rather than bulk-marked; and all prior live review threads are resolved. Exact-head repository CI and DCO are green. I found no remaining blocker in the current diff.

PR #601, opened 2026-09-03
fix(cli): lower rules on exposed prototype-owned state

### LINKED ISSUE #600: bug(cli): a Base-only token closure has no CSS for rules on exposed prototype-owned state
Found while reviewing the coverage gate in #585. A rule conditioned on a state the prototype owns and exposes lowers at runtime but has no variant in a closure generated from Base alone.

Baseline `40ba5fb9`.

## The gap

`packages/prototypes/base/src/tabs/content.proto.ts` creates its own state, exposes it, and styles on it:

```ts
const hidden = def.state.bool('hidden', true);
def.expose.state('hidden', hidden);
def.rule({ when: (w) => w.state(hidden).eq(true), intent: (i) => i.feedback.style.use(tw('hidden')) });
```

The Web runtime lowers that to `data-[hidden]:hidden` and drops the runtime fallback. The extractor derives a semantic only from `def.state.fromInteraction` and `def.state.fromAccessibility`, so a `def.state.bool` handle has none and the rule contributes the **bare** token instead:

```
collectProtoStyleTokens('packages/prototypes/base')
  closure size: 14
  bare `hidden`:            true
  `data-[hidden]:hidden`:   false
```

So the runtime applies a variant with no matching CSS, and inactive Base Tabs content is not hidden — `P-BASE-TABS-CONTENT-HIDDEN-WHEN-INACTIVE`.

## Why the shipped presets are fine

Both published closures contain `data-[hidden]:hidden`, but neither gets it from Base. `packages/prototypes/shadcn/src/tabs/content.proto.ts` and the Brutalist equivalent each write the same rule through `asTabsContent()`, whose state handles the extractor does resolve. The Base rule is masked by its projections rather than working.

Any consumer generating a closure from Base alone, or from a design language that does not restate this rule, gets the bare token and no variant.

## Suggested fix

`def.expose.state(name, handle)` is what gives a prototype-owned state a host attribute, so it is also what tells the extractor the variant is `data-[name]`. Registering that during the walk — the expose call precedes the rule in every current prototype — makes the Base rule lower on its own.

I prototyped it and confirmed it closes this case. I did **not** put it in #585: that PR aligns the coverage gate with the extractor and is test-only, while this changes generated CSS. The shadcn and Brutalist manifests would need regenerating and the diff reviewed on its own merits.

## Related

The gate in #585 currently skips prototype-owned states rather than reporting them, which is why this was invisible to it. Once the extractor derives the variant, the gate can treat an exposed local state as a real usage instead of a blind spot — worth doing in the same PR as the fix so the two stay aligned.

Happy to take it if you want it bounded that way.

HyacinthHaru:
Fixes #600 under the bounded authorization there. Follows the five recommended steps; adds no Host Capability and no spec semantic.

## The fix

`def.expose.state(name, handle)` is what gives a prototype-owned state a host attribute, so it is also what tells the extractor the selector. `registerExposedState` runs on the statements of each scope in both walks — the expose call precedes the rule in every current prototype — and gives an otherwise semantic-less handle `data-[<key>]`.

**Precedence matches the runtime, not the key.** `createExposeStateWebNameMap` maps an official semantic first and only falls back to the exposed key, so a handle that already carries one keeps it. A `def.state.fromInteraction('hovered')` exposed as `isHot` still lowers to `hover:` and never `data-[is-hot]:`.

**Normalization matches too.** The runtime kebab-cases an unannotated key, so `dragOver` becomes `data-drag-over`, and the extractor now applies the same transform rather than the raw key.

## Evidence

Base-only closure, before and after:

```
collectProtoStyleTokens('packages/prototypes/base')
  before:  size 14, bare `hidden` true, `data-[hidden]:hidden` false
  after:   size 15, bare `hidden` true, `data-[hidden]:hidden` true
```

The bare token stays, as the issue directed — nothing here justifies removing that over-approximation.

**Both preset manifests are byte-identical**: `check:styles:preset` reports 246 and 234 tokens, unchanged. That is the outcome the ruling anticipated, because the shadcn and Brutalist Tabs Content each already contribute `data-[hidden]:hidden` through `asTabsContent()`. The Base rule was masked by its projections rather than working, which is exactly why nothing was visibly broken.

Three extractor regressions: the Tabs-shaped bool fallback, the `dragOver` → `data-[drag-over]` normalization, and the precedence case. Mutation-checked by removing the registration again:

```
AssertionError: expected [ 'hidden' ] to include 'data-[hidden]:hidden'
```

## The gate side

Step 3 of the ruling. `scanRuleStateReads` no longer blanket-skips a prototype-owned state. An **exposed** one is reported in a new `exposedLocals` result rather than pretended to be an `asHook` pair; an unexposed one is still skipped, because neither analyzer lowers it.

The repository-wide gate now asserts that every exposed state a rule reads carries a key that can become a data attribute, and that the set is non-empty — so this path cannot quietly stop finding anything. It currently finds the Tabs Content case.

`packages` 1531 passed, `check:types` and `check:styles:preset` clean.

### DIFF
--- a/packages/cli/src/services/lowered-hook-coverage.ts
+++ b/packages/cli/src/services/lowered-hook-coverage.ts
@@ -310,16 +310,22 @@ export function scanRuleStateReads(
     return found.flatMap((edge) => aliasRoots(edge.target, chain, edge.at, next));
   };
 
+  // An exposure names a binding, and a binding lives in one scope. Writing it
+  // across the whole chain would let a sibling prototype reusing the same local
+  // name inherit an exposure its own runtime never registers.
+  const declaringScope = (name: string, chain: ts.Node[]): ts.Node | undefined =>
+    chain.find((candidate) => declaredIn.get(candidate)?.has(name)) ?? chain[chain.length - 1];
+
   for (const { handle, key, chain, at } of rawExposures) {
     // A key the scanner cannot read still means the state is exposed, and the
     // attribute comes from the declared name anyway.
     const text = ts.isStringLiteralLike(key) ? key.text : '';
     for (const name of new Set([handle, ...aliasRoots(handle, chain, at)])) {
-      for (const owner of chain) {
-        const scoped = exposedKeys.get(owner) ?? new Map<string, string>();
-        if (!scoped.has(name)) scoped.set(name, text);
-        exposedKeys.set(owner, scoped);
-      }
+      const owner = declaringScope(name, chain);
+      if (!owner) continue;
+      const scoped = exposedKeys.get(owner) ?? new Map<string, string>();
+      if (!scoped.has(name)) scoped.set(name, text);
+      exposedKeys.set(owner, scoped);
     }
   }
 
--- a/packages/cli/src/services/lowered-hook-coverage.ts
+++ b/packages/cli/src/services/lowered-hook-coverage.ts
@@ -216,24 +216,28 @@ export function scanRuleStateReads(
   const aliasRoot = (name: string, chain: ts.Node[], at: number): string => {
     const seen = new Set<string>();
     let current = name;
+    // Each hop resolves where that edge was created, not where the exposure was
+    // written, so a later redeclaration of an intermediate name cannot retarget
+    // an alias that had already captured its value.
+    let cursor = at;
     for (;;) {
       if (seen.has(current)) return current;
       seen.add(current);
-      // The edge in effect where the exposure was written, nearest scope first.
-      let next: string | undefined;
+      let found: AliasEdge | null = null;
       for (const owner of chain) {
         let best: AliasEdge | null = null;
         for (const edge of aliasEdges) {
-          if (edge.owner !== owner || edge.name !== current || edge.at > at) continue;
+          if (edge.owner !== owner || edge.name !== current || edge.at > cursor) continue;
           if (!best || edge.at > best.at) best = edge;
         }
         if (best) {
-          next = best.target;
+          found = best;
           break;
         }
       }
-      if (next === undefined) return current;
-      current = next;
+      if (!found) return current;
+      current = found.target;
+      cursor = found.at;
     }
   };
 
--- a/packages/cli/src/services/lowered-hook-coverage.ts
+++ b/packages/cli/src/services/lowered-hook-coverage.ts
@@ -153,6 +153,7 @@ export function scanRuleStateReads(
   const exposedKeys = new Map<ts.Node, Map<string, string>>();
   type AliasEdge = { owner: ts.Node; name: string; target: string; at: number };
   const aliasEdges: AliasEdge[] = [];
+  const declaredIn = new Map<ts.Node, Set<string>>();
   const rawExposures: Array<{
     handle: string;
     key: ts.Expression;
@@ -179,15 +180,21 @@ export function scanRuleStateReads(
 
   const collectExposedKeys = (node: ts.Node, chain: ts.Node[]): void => {
     const nextChain = introducesScope(node) ? [node, ...chain] : chain;
-    if (ts.isVariableDeclaration(node) && ts.isIdentifier(node.name) && node.initializer) {
-      const initializer = unwrap(node.initializer);
-      if (ts.isIdentifier(initializer)) {
-        aliasEdges.push({
-          owner: nextChain[0] ?? source,
-          name: node.name.text,
-          target: initializer.text,
-          at: node.getStart(),
-        });
+    if (ts.isVariableDeclaration(node) && ts.isIdentifier(node.name)) {
+      const owner = nextChain[0] ?? source;
+      const names = declaredIn.get(owner) ?? new Set<string>();
+      names.add(node.name.text);
+      declaredIn.set(owner, names);
+      if (node.initializer) {
+        const initializer = unwrap(node.initializer);
+        if (ts.isIdentifier(initializer)) {
+          aliasEdges.push({
+            owner,
+            name: node.name.text,
+            target: initializer.text,
+            at: node.getStart(),
+          });
+        }
       }
     }
     // A plain reassignment moves the handle just as a declaration does.
@@ -198,12 +205,15 @@ export function scanRuleStateReads(
     ) {
       const value = unwrap(node.right);
       if (ts.isIdentifier(value)) {
-        aliasEdges.push({
-          owner: nextChain[0] ?? source,
-          name: node.left.text,
-          target: value.text,
-          at: node.getStart(),
-        });
+        // An assignment does not declare, so it belongs to whichever scope owns
+        // the binding; otherwise a reassignment inside a nested block would be
+        // invisible to an exposure written outside it.
+        const left = node.left.text;
+        const owner =
+          [...nextChain, source].find((candidate) => declaredIn.get(candidate)?.has(left)) ??
+          nextChain[0] ??
+          source;
+        aliasEdges.push({ owner, name: left, target: value.text, at: node.getStart() });
       }
     }
     if (ts.isCallExpression(node)) {
--- a/packages/cli/src/services/lowered-hook-coverage.ts
+++ b/packages/cli/src/services/lowered-hook-coverage.ts
@@ -154,6 +154,32 @@ export function scanRuleStateReads(
   type AliasEdge = { owner: ts.Node; name: string; target: string; at: number };
   const aliasEdges: AliasEdge[] = [];
   const declaredIn = new Map<ts.Node, Set<string>>();
+  const objectMembers = new Map<string, Map<string, string>>();
+
+  /** Every handle an initializer may name; a conditional contributes both. */
+  const aliasTargets = (node: ts.Node): string[] => {
+    const value = unwrap(node);
+    if (ts.isIdentifier(value)) return [value.text];
+    if (ts.isConditionalExpression(value)) {
+      return [...aliasTargets(value.whenTrue), ...aliasTargets(value.whenFalse)];
+    }
+    return [];
+  };
+
+  const resolveHandleIdentifier = (node: ts.Node): string | null => {
+    const value = unwrap(node);
+    if (ts.isIdentifier(value)) return value.text;
+    const owner = ownerOf(value);
+    if (!owner) return null;
+    const base = unwrap(owner);
+    if (!ts.isIdentifier(base)) return null;
+    const members = objectMembers.get(base.text);
+    if (!members) return null;
+    for (const [key, target] of members) {
+      if (memberNamed(value, key)) return target;
+    }
+    return null;
+  };
   const rawExposures: Array<{
     handle: string;
     key: ts.Expression;
@@ -181,19 +207,41 @@ export function scanRuleStateReads(
   const collectExposedKeys = (node: ts.Node, chain: ts.Node[]): void => {
     const nextChain = introducesScope(node) ? [node, ...chain] : chain;
     if (ts.isVariableDeclaration(node) && ts.isIdentifier(node.name)) {
-      const owner = nextChain[0] ?? source;
+      // `var` binds to the enclosing function however deeply it is nested.
+      const isVar =
+        ts.isVariableDeclarationList(node.parent) &&
+        !(node.parent.flags & (ts.NodeFlags.Let | ts.NodeFlags.Const));
+      const enclosingFunction = isVar
+        ? ts.findAncestor(node, (candidate): candidate is ts.FunctionLikeDeclaration =>
+            ts.isFunctionLike(candidate)
+          )
+        : undefined;
+      const enclosingBody = enclosingFunction?.body;
+      const owner =
+        (enclosingBody && nextChain.includes(enclosingBody) ? enclosingBody : nextChain[0]) ??
+        source;
       const names = declaredIn.get(owner) ?? new Set<string>();
       names.add(node.name.text);
       declaredIn.set(owner, names);
       if (node.initializer) {
-        const initializer = unwrap(node.initializer);
-        if (ts.isIdentifier(initializer)) {
-          aliasEdges.push({
-            owner,
-            name: node.name.text,
-            target: initializer.text,
-            at: node.getStart(),
-          });
+        const literal = unwrap(node.initializer);
+        if (ts.isObjectLiteralExpression(literal)) {
+          const members = new Map<string, string>();
+          for (const property of literal.properties) {
+            if (ts.isShorthandPropertyAssignment(property)) {
+              members.set(property.name.text, property.name.text);
+            } else if (
+              ts.isPropertyAssignment(property) &&
+              (ts.isIdentifier(property.name) || ts.isStringLiteralLike(property.name))
+            ) {
+              const value = unwrap(property.initializer);
+              if (ts.isIdentifier(value)) members.set(property.name.text, value.text);
+            }
+          }
+          objectMembers.set(node.name.text, members);
+        }
+        for (const target of aliasTargets(node.initializer)) {
+          aliasEdges.push({ owner, name: node.name.text, target, at: node.getStart() });
         }
       }
     }
@@ -203,17 +251,16 @@ export function scanRuleStateReads(
       node.operatorToken.kind === ts.SyntaxKind.EqualsToken &&
       ts.isIdentifier(node.left)
     ) {
-      const value = unwrap(node.right);
-      if (ts.isIdentifier(value)) {
-        // An assignment does not declare, so it belongs to whichever scope owns
-        // the binding; otherwise a reassignment inside a nested block would be
-        // invisible to an exposure written outside it.
-        const left = node.left.text;
-        const owner =
-          [...nextChain, source].find((candidate) => declaredIn.get(candidate)?.has(left)) ??
-          nextChain[0] ??
-          source;
-        aliasEdges.push({ owner, name: left, target: value.text, at: node.getStart() });
+      // An assignment does not declare, so it belongs to whichever scope owns
+      // the binding; otherwise a reassignment inside a nested block would be
+      // invisible to an exposure written outside it.
+      const left = node.left.text;
+      const assignOwner =
+        [...nextChain, source].find((candidate) => declaredIn.get(candidate)?.has(left)) ??
+        nextChain[0] ??
+        source;
+      for (const target of aliasTargets(node.right)) {
+        aliasEdges.push({ owner: assignOwner, name: left, target, at: node.getStart() });
       }
     }
     if (ts.isCallExpression(node)) {
@@ -224,10 +271,10 @@ export function scanRuleStateReads(
       const exposesDirectly = memberNamed(callee, 'expose');
       if (exposesState || exposesDirectly) {
         const [nameArg, handleArg] = node.arguments;
-        const handle = handleArg && unwrap(handleArg);
-        if (nameArg && handle && ts.isIdentifier(handle)) {
+        const handle = handleArg && resolveHandleIdentifier(handleArg);
+        if (nameArg && handle) {
           rawExposures.push({
-            handle: handle.text,
+            handle,
             key: nameArg,
             chain: [...nextChain, source],
             at: node.getStart(),
@@ -239,39 +286,35 @@ export function scanRuleStateReads(
   };
   collectExposedKeys(source, []);
 
-  const aliasRoot = (name: string, chain: ts.Node[], at: number): string => {
-    const seen = new Set<string>();
-    let current = name;
-    // Each hop resolves where that edge was created, not where the exposure was
-    // written, so a later redeclaration of an intermediate name cannot retarget
-    // an alias that had already captured its value.
-    let cursor = at;
-    for (;;) {
-      if (seen.has(current)) return current;
-      seen.add(current);
-      let found: AliasEdge | null = null;
-      for (const owner of chain) {
-        let best: AliasEdge | null = null;
-        for (const edge of aliasEdges) {
-          if (edge.owner !== owner || edge.name !== current || edge.at > cursor) continue;
-          if (!best || edge.at > best.at) best = edge;
-        }
-        if (best) {
-          found = best;
-          break;
-        }
-      }
-      if (!found) return current;
-      current = found.target;
-      cursor = found.at;
+  // Fans out rather than picking one edge, so a conditional alias gives every
+  // candidate handle its variant.
+  const aliasRoots = (
+    name: string,
+    chain: ts.Node[],
+    at: number,
+    seen = new Set<string>()
+  ): string[] => {
+    if (seen.has(name)) return [name];
+    let found: AliasEdge[] = [];
+    for (const owner of chain) {
+      const candidates = aliasEdges.filter(
+        (edge) => edge.owner === owner && edge.name === name && edge.at <= at
+      );
+      if (candidates.length === 0) continue;
+      const latest = Math.max(...candidates.map((edge) => edge.at));
+      found = candidates.filter((edge) => edge.at === latest);
+      break;
     }
+    if (found.length === 0) return [name];
+    const next = new Set([...seen, name]);
+    return found.flatMap((edge) => aliasRoots(edge.target, chain, edge.at, next));
   };
 
   for (const { handle, key, chain, at } of rawExposures) {
     // A key the scanner cannot read still means the state is exposed, and the
     // attribute comes from the declared name anyway.
     const text = ts.isStringLiteralLike(key) ? key.text : '';
-    for (const name of new Set([handle, aliasRoot(handle, chain, at)])) {
+    for (const name of new Set([handle, ...aliasRoots(handle, chain, at)])) {
       for (const owner of chain) {
         const scoped = exposedKeys.get(owner) ?? new Map<string, string>();
         if (!scoped.has(name)) scoped.set(name, text);
--- a/packages/cli/src/services/lowered-hook-coverage.ts
+++ b/packages/cli/src/services/lowered-hook-coverage.ts
@@ -102,6 +102,18 @@ function lookup(scope: Scope | null, name: string): Binding | null {
   return null;
 }
 
+function isLoopStatement(node: ts.Node): boolean {
+  return ts.isForStatement(node) || ts.isForOfStatement(node) || ts.isForInStatement(node);
+}
+
+/** The scope a name is already bound in, so an assignment does not shadow it. */
+function scopeBinding(scope: Scope, name: string): Scope | null {
+  for (let current: Scope | null = scope; current; current = current.parent) {
+    if (current.bindings.has(name)) return current;
+  }
+  return null;
+}
+
 function introducesScope(node: ts.Node): boolean {
   return (
     ts.isSourceFile(node) ||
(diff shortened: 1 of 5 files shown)


cyjin-yl (changes requested):
Reviewed exact head `c04863cf120a4140107ae1fcdc21015a6ae6b9f1`. The earlier expose-key/ordering/constant/element-access/redeclaration findings are materially fixed, and the current repository CI is green, but this head still has three extractor/runtime alignment blockers:

1. `collectExposures()` is component-wide in intent but file-wide in implementation: it builds one `aliases: Map<string,string>` keyed only by identifier spelling for the entire source. Sibling prototype/setup scopes that reuse the same alias name can overwrite each other's alias edge, so an exposure in prototype A can be attributed to prototype B's handle. The scanner mirrors the same global-name model, so the coverage gate can stay green. Track lexical declaration identity / state-handle identity (or otherwise isolate component scopes) and add a regression with two sibling setup scopes reusing the same alias spelling.

2. `resolveDeclaredStateName()` correctly resolves literals and static constants, but when a valid state name comes from an opaque-but-runtime-stable expression it returns null and `applyExposure()` falls back to the public expose key. That is not the runtime fallback: `StateKernel.define()` always stores the runtime value of the declaration name in `__stateSemantic`, and the Web expose path prefers that semantic over the key. Emitting `data-[<expose-key>]` here therefore certifies the wrong selector. If the declared semantic cannot be proven statically, fail closed/report it as unresolved rather than substituting the expose key; add a regression for an unresolved declaration-name expression exposed under a different key.

3. Exposed `numberDiscrete` states still cannot be lowered for numeric equality. The current `resolveStateEqVariant()` accepts booleans and string literals only, while the Web rule-expose implementation explicitly accepts `number.discrete` and stringifies its literal (`data-[name=1]`). A rule such as `w.state(step).eq(1)` can therefore be removed by the runtime optimizer while the CLI emits no matching variant CSS. Preserve enough state-kind information or accept the supported numeric literal shape, and cover the exposed discrete-number case in both extraction and coverage-gate regressions.

Acceptance: the production extractor and coverage scanner must agree with the Web runtime for all three cases, with focused regressions demonstrating that none can pass the repository gate while omitting/misnaming the generated variant.

cyjin-yl (changes requested):
Re-reviewed exact head `88a1925a3e5945d2adca5e7cbbecbc288e31490b` after the new fix commit. The prior three production findings are materially improved: sibling setup functions no longer share one alias map, an opaque declaration name no longer falls back to the expose key in the extractor, and numeric `number.discrete` equality now produces a data variant. Three current-head alignment blockers remain:

1. The coverage scanner still uses one file-wide `aliasOf` / `exposedKeys` map keyed only by identifier spelling. The production extractor is now function-scoped, so two sibling prototypes that both declare `publicFlag` can be analyzed differently by the gate and extractor; the gate can still certify the wrong exposed handle. Please make exposure/alias tracking lexical (or share the production resolver/state identity) and add the sibling-setup case to `scanRuleStateReads`/the gate, not only to the extractor test.

2. The scanner still records a local state's `declaredAs` only when the state name is a string literal, then computes `attribute` as `declaredAs ?? exposedAs`. For `def.state.bool(makeStateName(), false)` exposed as `visible`, the extractor now correctly emits no variant because the runtime semantic cannot be proven, but the scanner reports `data-visible` as a valid exposed local and the repository gate stays green. This is the exact fail-closed mismatch the previous review asked to eliminate. If the declared runtime semantic is not statically resolvable, report the rule/exposure as unresolved (or use the same resolver result) rather than substituting the expose key; cover the opaque-name case in the gate regression as well.

3. `collectExposures()` scopes alias edges by function, but not by lexical block. Two sibling blocks inside one `setup` may legally reuse the same `const publicFlag` name for different state handles. Both edges still land in the same function map, the later declaration overwrites the earlier one, and the first exposure can be attributed to the wrong handle. Track declaration/scope identity through block scopes (the normal extractor scope model already distinguishes them) and add a regression with sibling blocks reusing the alias spelling.

Acceptance remains that production extraction and the repository coverage gate agree for every exposed-local rule shape they claim to cover. Exact-head CI `33721810129` is also still in progress at review time, so this head would not be mergeable yet even after the semantic blockers above.

cyjin-yl (changes requested):
Re-reviewed exact head `d2e8436a38cea5b0a2a4bee6aaee0ded9b93b40f`. The three blockers from my prior review are materially addressed: exposure tracking is now lexical across sibling/block scopes, opaque declaration names fail closed instead of substituting the expose key, and exposed discrete-number comparisons are lowered. Two current-head extractor/runtime alignment blockers remain:

1. `collectExposures()` still resolves a multi-hop alias chain using the exposure position for every hop. In `var current = first; const publicFlag = current; var current = second; def.expose.state('visible', publicFlag)`, `publicFlag` captured `first` when it was initialized, but `rootName(publicFlag, ..., exposureAt)` resolves `publicFlag -> current` and then looks up `current` again at the later exposure position, selecting `second`. The Web runtime captured the actual handle value (`first`), so a rule on `first` can be optimized away at runtime while the CLI omits/misattributes the variant. `lowered-hook-coverage.ts` uses the same exposure-time walk in `aliasRoot`, so the gate can agree with the wrong static result. Preserve target identity (or the target lookup time/declaration identity) at each alias edge, and add matching extractor + coverage regressions for a multi-hop alias whose intermediate target is redeclared after the alias was created.

2. `signedNumericText()` reconstructs signed source spelling rather than the runtime literal value. For `w.state(step).eq(-0)`, it emits `-0`, producing `data-[step=-0]`; the runtime's numeric literal is JavaScript negative zero and `String(condition.literal)` is `"0"`, so the host projects `data-step="0"` and the generated selector never matches. Serialize the evaluated numeric value/canonical runtime representation (at minimum `-0 -> 0`) and cover the exact result in extraction; the coverage path must not certify this shape unless it agrees with the same canonical value semantics.

Both findings also have current, non-outdated unresolved review threads on this head. Acceptance is that production extraction and the coverage gate match the Web runtime for these two cases, with focused mutation-resistant regressions.

Separately, trusted repository CI run `33724810172` is red in the real `test` job (`Public documentation gate and repository tests`). There is now evidence of an independent main-branch Tooltip browser-test defect in #603 that may account for that failure, but merge policy still requires a successful trusted repository CI result for this exact head after the repository-side failure is corrected/reconciled. Vercel fork authorization is not being treated as a code failure.

cyjin-yl (changes requested):
Re-reviewed exact head `4634f6b2b41036ecddb7b984a34ee3b5cd043429`. The two blockers from my prior review are materially fixed: alias chains now resolve each hop at the edge where it was created, and numeric lowering canonicalizes `-0` to the runtime spelling `0`. The follow-up fixes also cover plain reassignments, nested-block assignment ownership, and transparent state initializers. One current-head extractor/gate identity blocker remains.

`collectExposures(sourceFile)` ultimately flattens all discovered exposures into one `byHandle: Map` keyed only by identifier spelling, and `applyExposure()` later consults it with `exposures.get(name)` for every declaration in the file. That loses the lexical/component identity which the alias pre-pass just worked to preserve. A sibling prototype that happens to reuse the same original local name is therefore contaminated by another prototype's exposure. For example, prototype A can declare and expose `const flag = def.state.bool('firstFlag', ...)`, while prototype B independently declares an **unexposed** `const flag = def.state.bool('secondFlag', ...)` and has a rule on B's `flag`; the source-wide `byHandle` entry for `"flag"` makes B's declaration appear exposed and the CLI may lower its rule to `data-[second-flag]:...` even though B's runtime never exposes that state. The current sibling regression only collides the alias spelling (`publicFlag`) while the original handles are named `flag` and `other`, so it does not exercise this case.

The coverage scanner has the same false-positive path: when materializing a raw exposure it copies the handle spelling into every owner in the exposure chain, including the source scope, and `exposureFor()` for a sibling setup can therefore find the source-level `flag`. The repository gate can agree with the same wrong result rather than catching the production mismatch.

Please keep exposure identity scoped to the declaration/component that owns the state (or share a declaration-identity resolver between extractor and scanner) instead of collapsing it to a source-wide identifier string. Add matching extractor + coverage regressions with two sibling prototypes/setups that both use the same original state variable name, only one of which is exposed; the unexposed sibling must remain unexposed and must not be certified/lowered by the gate.

Separately, exact-head trusted CI run `33729877049` is still red in the real `test` job (`Public documentation gate and repository tests`), although DCO, type-check, lockstep, release-scan and the other applicable lanes are green. #604 has now merged the independent Tooltip browser-test repair that was causing this class of failure, so this PR needs a successful trusted repository CI result on its live exact head/merge candidate after that base change. GitHub also currently reports the PR as not mergeable against the moved `main`, so the branch/base conflict must be reconciled and the resulting head re-reviewed if it changes. Vercel fork authorization is not being treated as a code failure.

cyjin-yl (changes requested):
Re-reviewed exact head `50d8f96faa135d0d4e458d58b7c7c77fec1e0ec4`. The prior source-wide exposure-identity blocker is materially fixed: exposure records are now keyed by lexical declaration owner rather than one file-wide handle-name map, and trusted repository CI run `33732566994` is green. Four current-head extractor/gate alignment blockers remain:

1. `prototype-style-tokens.ts` still derives exposed local-state attributes with generic `exposedDataAttributeName()` only. The runtime `createExposeStateWebNameMap()` first applies the official semantic map (`@accessibility/checked -> checked`, etc.). An exposed local declared as `@accessibility/checked` therefore produces static `data-[accessibility-checked]` while runtime lowers to `data-[checked]` and removes the runtime rule. Reuse/share the official semantic mapping before generic normalization, and cover the extractor + gate with an official-semantic local-state regression.

2. `collectExposures()` keeps `objectMembers` in one file-wide map keyed only by identifier spelling. With outer `controls = { ready: first }`, a nested shadowing `controls = { ready: second }`, and a later outer `def.expose.state(..., controls.ready)`, the nested object can overwrite the lookup and attribute the exposure to the wrong handle. Runtime lexical lookup still sees the outer object. Scope property-backed handle bindings by lexical declaration/source position (or declaration identity), and add matching extractor + scanner regressions for object shadowing.

3. `lowered-hook-coverage.ts` still records a local state's `declaredAs` only when the declaration argument is a string literal. Production extraction now resolves statically known constants, so `const name = 'hidden'; const flag = def.state.bool(name, false)` is extractable but the gate reports it unresolved. Use equivalent constant-expression resolution in the scanner (or share the resolver) and add a regression proving valid constant-backed names remain accepted.

4. The production walk still registers declarations only for `VariableStatement`s that are direct members of a statement-bearing scope. A declaration under single-statement control flow, e.g. `if (true) var flag = def.state.bool('flag', false);`, is recursively visited but never registered, so a later exposed-state rule can lose its variant even though runtime executed the declaration; the scanner does register the nested declaration, allowing a green mismatch. Register nested declarations with the correct `var`/lexical ownership semantics, or fail closed, and cover this shape in production + gate tests.

These findings correspond to the current non-outdated unresolved review threads. Acceptance is that production extraction and the repository coverage gate agree with the Web runtime for all four cases, with focused mutation-resistant regressions. CI is no longer a blocker on this head, but the semantic blockers and unresolved threads are, so this head is not ready to approve or merge.

cyjin-yl (changes requested):
Re-reviewed exact head `9c43a9fca09ffe03ee31f55744f966f25e69176f`. The four blockers from my prior review are materially addressed: official exposed-state names are mapped before normalization, property-backed handles are now scoped/positioned, constant-backed declaration names are understood by the scanner, and single-statement `VariableStatement`s are registered. Exact-head repository CI (`33744529382`) is green. Four current-head extractor/runtime alignment blockers remain, matching the unresolved non-outdated threads:

1. Owned official interaction states lose the data-attribute fallback when `applyExposure()` replaces their semantic with `hover` / `active`. For `eq(false)`, `resolveStateEqVariant()` can only negate a `data-[…]` semantic, while the runtime optimizer only uses the native variant for `true` and falls back to `not-[data-hovered]` / `not-[data-pressed]` for `false`. In an `all(...)` with another positive condition, runtime therefore removes the rule using the full selector while the CLI can silently drop the negative condition and emit an over-broad selector. Preserve both the native-positive semantic and the data fallback (or model the runtime decision directly), with a mixed positive + negated hovered/pressed regression in extractor and gate.

2. `walk()` still does not register lexical loop initializers. A `for (let flag = def.state.bool('inner', ...); ... )` initializer is a `VariableDeclarationList`, not a `VariableStatement`; if it shadows an outer `flag`, the rule body resolves the outer binding while runtime uses/exposes the inner handle. Add proper loop/catch lexical-scope declaration registration (or fail closed) and a shadowing regression shared with the coverage path.

3. The exposure prepass follows ordinary alias reassignments, but the production expression scope does not update its binding on assignment. `let current = first; current = second; expose(current); rule(w.state(current))` can therefore expose/lower `second` at runtime while the rule analyzer still reads `first`. Keep ordinary binding state in sync with statically traceable assignments, or conservatively refuse this shape, and cover the alias-read-after-reassignment case in both paths.

4. Destructured state-handle aliases are still absent from exposure identity tracking. `const { ready: publicFlag } = { ready: flag }; def.expose.state('visible', publicFlag)` is a valid runtime alias of `flag`, but `collectExposures()` only records alias edges for identifier declaration names, and the scanner mirrors that restriction. The runtime can remove a rule on `flag` while generated CSS lacks the required variant. Trace statically resolvable object/array binding patterns or fail closed, with extractor + gate regressions.

Acceptance is unchanged: every exposed-local rule shape the repository gate certifies must agree with the Web optimizer on the exact selector and must not allow runtime rule removal without matching generated CSS. CI is no longer a blocker on this head; the four semantic mismatches and their unresolved threads are.

cyjin-yl (changes requested):
Reviewed exact head `e7a4ad004f6519c9ccd1bfac9c961a5f14e8f67a`. The four blockers from my previous review are materially addressed, and the trusted repository CI run `33749172791` is green; the fork Vercel authorization status is not being treated as a code failure. I still cannot approve this head because five current-head extractor/gate mismatches remain:

1. `packages/cli/src/services/prototype-style-tokens.ts`: member-backed handle resolution drops the selected member edge's source position. In `let current = first; const controls = { ready: current }; current = second; def.expose.state('visible', controls.ready)`, the object still contains `first`, but resolving the stored target name at the later exposure position retargets it to `second`. Preserve the member edge's `at` when following its target, and add matching extractor + coverage regressions.
2. The alias prepass treats a write under conditional control flow as unconditional. `let current = first; if (enabled) current = second; expose(current)` must conservatively retain both possible state identities unless the branch is statically known; otherwise one runtime path loses its generated variant while the gate stays green.
3. The sequential production walker still block-scopes nested `var` redeclarations even though JavaScript hoists them to the function binding. A nested `var current = second` followed by `expose(current); rule(current)` outside the block can therefore read `first` statically while runtime reads `second`. Make ordinary binding ownership match the exposure prepass/function-scoped `var` semantics and regress it.
4. `packages/cli/src/services/lowered-hook-coverage.ts` still rejects element-access state declarations such as `def.state['bool'](...)` even though the production extractor now accepts them. Reuse the same static member-name recognition so valid extractable code does not fail the gate.
5. The coverage assertion still requires a normalized data-attribute suffix to begin with a letter. A valid stable state name like `1st` lowers to `data-1st` / `data-[1st]` in the runtime and production extractor, so the gate must validate the generated attribute/selector rather than impose an identifier-only naming rule absent from the contract.

These are not speculative style nits: the first three can make runtime-lowered rules and generated CSS disagree while the coverage gate remains green; the last two make the gate reject code the production path supports. Please keep the scanner and production extractor aligned and add negative/positive regressions for each case. The five corresponding current-head review threads are still unresolved.

cyjin-yl (changes requested):
Re-reviewed exact head `1a8fbd602c59af4d29a2df77d13e107d10d21495` after the author’s follow-up commit. The five blockers from my previous review are materially addressed, and trusted repository CI run `33773517047` is green, including the real `test`, `type-check`, `lockstep-check`, and `release-scan` lanes. I still cannot approve this head because the production extractor and coverage scanner can diverge from runtime state-handle identity in three current shapes:

1. **Conditional reassignment is still applied unconditionally by the sequential rule walk.** `collectExposures()` now correctly keeps both candidates for `let current = first; if (enabled) current = second; expose(current)`, but `walk()` always executes `owner.bindings.set(name, ...)` when it encounters `current = second`, regardless of whether that write is conditionally reached. If a later rule reads `current`, an `enabled === false` runtime uses/exposes `first` while the collector can emit only the selector for `second`. `lowered-hook-coverage.ts` mirrors the same unconditional reassignment in its ordinary binding walk, so the gate can agree with the wrong result. Preserve the union/conditional identity for subsequent reads, or fail closed for that rule shape, and add matching extractor + gate regressions for a rule that reads the alias after a conditional write.

2. **An exposed state read through an ordinary object member is still not resolvable in rule conditions.** The exposure prepass now understands `const controls = { ready: flag }; def.expose.state('visible', controls.ready)`, but `resolveStateHandleSemantic()` only gets member semantics from `resolveExpression(...).semanticMap`; ordinary object literals are represented as token-string maps and do not carry state-handle semantic identity. Therefore `w.state(controls.ready)` can remain unresolved while runtime reads the same state object and removes the rule in favor of its data variant. Carry state-handle identity through statically traceable object/member bindings in both production and the scanner, with a regression that the member expression itself lowers to the exact runtime selector.

3. **The same identity loss remains for destructured local-state aliases used by the rule.** The exposure prepass now traces patterns such as `const { ready: publicFlag } = { ready: flag }`, but the ordinary declaration path only destructures `semanticMap` values; an ordinary object containing a prototype-owned state does not provide that map. So `expose(publicFlag); rule(w.state(publicFlag))` can be recognized as an exposure in the prepass while the rule-side alias still lacks the state semantic. Please cover both object and array destructuring (or conservatively reject/fail closed) and prove the alias read—not only the original root handle—produces the required variant in both extractor and coverage-gate tests.

Acceptance remains: every exposed-local rule shape the repository gate certifies must resolve to the exact selector the Web optimizer uses, with no path where runtime rule removal can leave generated CSS without that variant. The first two findings also correspond to current non-outdated unresolved review threads on this exact head. CI is no longer a blocker; these semantic alignment issues are.

cyjin-yl (changes requested):
Re-reviewed exact head `7e89d74ba9e8271b6de8a72740cfebd606bbade5`. The three blockers from my previous review are materially addressed, and the trusted repository CI for this exact head is green. I still cannot approve because one rule-side container mutation gap remains.

**Blocker — member writes update exposure identity, but not the binding that `w.state(container.member)` reads.**

`packages/cli/src/services/prototype-style-tokens.ts` now tracks `controls.ready = second` inside `collectExposures()`, so an exposure after that write is correctly attributed to `second`. But the sequential `walk()` only updates ordinary bindings for `Identifier = ...`; property/element assignments never update the `semanticMap` stored for `controls`. `resolveExpression(controls.ready)` therefore continues to read the object literal's declaration-time member.

Concrete failing shape:

```ts
const first = def.state.bool('firstFlag', false);
const second = def.state.bool('secondFlag', false);
const controls = { ready: first };
controls.ready = second;
def.expose.state('visible', controls.ready);
def.rule({
  when: (w) => w.state(controls.ready).eq(true),
  intent: (i) => i.feedback.style.use(tw('bg-accent')),
});
```

At runtime both the exposure and the rule read `second`, so the Web optimizer can remove the rule in favor of `data-[second-flag]`. The static exposure prepass also sees `second`, but the ordinary rule walk still sees the stale declaration-time `first` member (which is not exposed here), so the required variant can be omitted. The coverage scanner has the same split: its exposure prepass records member writes, while `readState()` / `containerMember()` resolves `controls.ready` from the original token initializer and can classify the stale `first` as an ordinary unexposed local, leaving the gate green.

The current regression `follows a handle written into the container before the exposure` does not cover this: its rule reads `second` directly rather than `controls.ready`.

**Fix/verification:** keep rule-side container member identity synchronized with statically named property/element writes (including conditional writes conservatively), or fail closed when the current member cannot be proven. Add matching extractor + coverage regressions where the rule itself reads `controls.ready` after the write, and prove the emitted selector matches the Web runtime. Please include the analogous statically named element-access write/read if that path is supported by the exposure prepass.

No other new blocker found in the current increment.

cyjin-yl (changes requested):
Re-reviewed exact head `e739ee80c3c9c5cb58298086e2bd9d020b0aa25d`. The previous direct-member-write blocker is materially fixed: both `prototype-style-tokens.ts` and `lowered-hook-coverage.ts` now update a statically named container member for subsequent rule reads, including conservative retention across conditional writes, and this head's trusted CI is green. One current-head identity blocker remains.

**Blocker — writes through an alias of the container do not update the original container's member identity.**

The new production branch handles `controls.ready = second` by cloning `controls.semanticMap`, updating that member, and rebinding only the identifier that appears on the left-hand side. But ordinary object aliases are JavaScript reference aliases. For example:

```ts
const first = def.state.bool('firstFlag', false);
const second = def.state.bool('secondFlag', false);
const controls = { ready: first };
const alias = controls;
alias.ready = second;
def.expose.state('visible', controls.ready);
def.rule({
  when: (w) => w.state(controls.ready).eq(true),
  intent: (i) => i.feedback.style.use(tw('bg-accent')),
});
```

At runtime `alias` and `controls` are the same object, so both the exposure and rule read `second`; the Web optimizer can remove the runtime rule in favor of `data-[second-flag]`. In the static walker, `alias` initially receives the same resolved container value, but the member-write path replaces only the `alias` binding with a new `{ ...container, semanticMap }`; the `controls` binding retains its declaration-time `first` member. The exposure prepass / coverage scanner likewise key member writes by the syntactic base identifier, so a write through `alias` does not retarget a later read through `controls`. The extractor and gate can therefore agree on the stale identity while runtime uses another handle.

Please preserve container/object identity across statically traceable aliases (including property/element reads and writes), or fail closed when an aliasing write makes the current member identity unprovable. Add matching production-extractor and coverage-gate regressions where the member is written through `alias` but both `def.expose.state(...)` and `w.state(...)` read through `controls`; the emitted selector must be the runtime selector for `second`. Conditional alias writes should conservatively retain every possible handle identity.

Exact-head CI run `33783849023` is otherwise successful (`test`, `type-check`, `lockstep-check`, `release-scan`, package/release lanes), docs preview is successful, and DCO reports all commits signed off. The separate Vercel fork-authorization failure is not being treated as a code failure. No other new blocker found in this increment.

cyjin-yl (changes requested):
Re-reviewed exact head `7a32fa3332c49e6242316570e1aa73f9a6b32da3`. The previous container-alias blocker is materially fixed: writes through `const alias = controls` now mutate the shared resolved container identity, and the current extractor/gate regressions cover the ordinary alias case. Trusted repository CI and DCO are green on this head; the fork Vercel authorization status is not a code failure. One current-head blocker remains.

**Blocker — `isImmediatelyInvoked()` treats async/generator IIFEs as synchronously completed writes.**

Both `packages/cli/src/services/prototype-style-tokens.ts` and `packages/cli/src/services/lowered-hook-coverage.ts` now decide that a write from another function is ordered whenever the enclosing function expression is directly called. That is sound for a normal synchronous IIFE only. An async IIFE can suspend before the write, and a generator IIFE does not execute its body at all merely by being called. For example:

```ts
const first = def.state.bool('firstFlag', false);
const second = def.state.bool('secondFlag', false);
let current = first;
(async () => {
  await 0;
  current = second;
})();
def.expose.state('visible', current);
def.rule({ when: (w) => w.state(current).eq(true), intent: use });
```

At setup/rule registration time the runtime still holds `first`; the continuation assigning `second` runs later. The current AST walk nevertheless visits that assignment immediately, `isImmediatelyInvoked()` returns true for the async arrow, and `isDeferredWrite()` therefore treats it as ordered. The collector can collapse the rule/exposure to `second` and omit the `data-[first-flag]` variant that the runtime optimizer uses. The coverage scanner mirrors the same classification, so the repository gate can stay green. A generator IIFE such as `(function* () { current = second; })()` is an even stricter instance: calling it only returns an iterator and never executes the assignment.

Fix criterion: only exempt an IIFE from deferred-write handling when its body is guaranteed to execute synchronously through the write. At minimum, async and generator functions must remain deferred/uncertain; if the analyzer cannot establish synchronous reachability, preserve the prior candidate or fail closed. Please add matching extractor and coverage regressions for an async IIFE with an `await` before the assignment (and preferably a generator IIFE) showing that the pre-write handle remains covered. This should stay an analysis/gate correction; it does not require changing runtime semantics or spec.

cyjin-yl (changes requested):
Re-reviewed exact head `450b7b87aa3db1c5fc3c9307a7958d2c0ad7872a` after the author’s follow-up. The async/generator IIFE blocker from my previous review is materially fixed: both analyzers now reject async and generator functions from the in-place-write exemption, and the new regressions cover those two shapes. Trusted repository CI, docs preview, and DCO are green on this head; the separate Vercel fork-authorization status is not being treated as a code failure.

One current-head soundness blocker remains in the new `runsInPlace()` criterion.

**Blocker — a synchronous direct IIFE is treated as if every write in its body necessarily executed.**

`packages/cli/src/services/prototype-style-tokens.ts` and `packages/cli/src/services/lowered-hook-coverage.ts` now classify a write as ordered whenever its enclosing function is non-async, non-generator, and directly invoked. That proves the function begins synchronously, but it does not prove the particular assignment was reached before the IIFE completed. The AST walk still visits unreachable statements after an abrupt completion, while `isConditionallyReached()` does not model `return`/`throw` reachability.

For example:

```ts
const first = def.state.bool('firstFlag', false);
const second = def.state.bool('secondFlag', false);
let current = first;
(() => {
  return;
  current = second;
})();
def.expose.state('visible', current);
def.rule({
  when: (w) => w.state(current).eq(true),
  intent: (i) => i.feedback.style.use(tw('bg-accent')),
});
```

At runtime the assignment never executes, so both exposure/rule registration still use `first`, and the Web optimizer may remove the runtime rule in favor of `data-[first-flag]`. The static walker nevertheless visits `current = second`; `runsInPlace()` returns true for the synchronous direct IIFE, `isDeferredWrite()` returns false, and the write replaces the prior candidate. `collectExposures()` uses the same classification, and the coverage scanner mirrors it, so extractor and gate can agree on `second` while omitting the selector the runtime actually uses.

Fix criterion: the exemption must be based on whether the specific write is guaranteed to have executed before the later read, not merely whether its enclosing function was synchronously called. Either add enough control-flow reachability to prove that guarantee, or conservatively keep the prior candidate/fail closed whenever earlier abrupt completion or otherwise-unproven reachability exists. Please add matching extractor + coverage regressions for a synchronous IIFE with an early `return` before the assignment (an early `throw` is the analogous case), while retaining the existing regression proving a definitely reached synchronous-IIFE write does replace the old handle.

No other new blocker found in this increment.

cyjin-yl (changes requested):
Re-reviewed exact head `55d0fb5d028a795bb438f9d16cb6b117a819769d`. The previous early-return/throw IIFE blocker is fixed: both analyzers now distinguish a synchronous IIFE starting from the specific write actually being reached, with matching regressions. The new ambiguous-base and unreadable-key handling is also conservative in the cases covered by the added tests, and trusted repository CI is green on this head.

There is still one blocking identity/timing gap in the production collector's member-write path. In `packages/cli/src/services/prototype-style-tokens.ts`, a write through a container alias computes deferred-ness from `scopeDeclaring(base.text, scope)`, i.e. the syntactic alias's declaring scope. That is not necessarily the scope/lifetime of the shared container object being mutated.

A concrete counterexample is:

```ts
const first = def.state.bool('firstFlag', false);
const second = def.state.bool('secondFlag', false);
const controls = { ready: first };

def.on('refresh', () => {
  const alias = controls;
  alias.ready = second;
});

def.expose.state('visible', controls.ready);
def.rule({
  when: (w) => w.state(controls.ready).eq(true),
  intent: (i) => i.feedback.style.use(tw('bg-accent')),
});
```

At setup/rule-registration time the callback has not run, so runtime `controls.ready` is still `first`; the Web lowering therefore needs the `data-[first-flag]` variant. Static resolution correctly knows `alias` and `controls` share one container value, but at `alias.ready = second` the alias itself is declared inside the callback. `scopeDeclaring('alias', scope)` therefore points at that callback, and `isDeferredWrite(node, enclosingFunction(declaring?.node))` sees the writing function as the same function and returns non-deferred. The shared `semanticMap` is then overwritten to `second`, so the later outer read through `controls.ready` can lose the `first` candidate even though that write has not happened at runtime.

The coverage scanner's later sequential member-write path is actually closer to the needed behavior: it follows `containerBindings(base, current)` to the underlying target binding and evaluates `isDeferredWrite` against `target.scope.node`. The production collector should preserve that same underlying-container lifetime/identity rather than using the callback-local alias declaration as the timing boundary.

Please fix member-write deferredness so writes through a local alias are measured against the shared/owning container identity (or otherwise conservatively retain the prior member candidate / fail closed). Add extractor and coverage-gate regressions where the alias is declared inside a deferred callback, the write is through that alias, and both exposure and rule later read the original container outside the callback. The required `first` variant must remain present; a definitely executed synchronous write should still be allowed to replace it.

Until that case is covered, the exact head can still produce a green gate while omitting a runtime-required selector, so I cannot approve it.

cyjin-yl (approved):
Reviewed exact head `49cdd746630c83fe6f844cda187a04e0ce2d3157` after the author’s follow-up. The previous callback-local container-alias timing blocker is materially fixed: the production member-write path now determines deferredness from the outermost scope that owns the shared resolved container value, so an alias declared inside `def.on(...)` no longer makes a deferred write look ordered against itself. The focused regressions cover both sides of the boundary: the callback-local alias retains both `first-flag` and `second-flag`, while the same write in a synchronously executed plain block replaces the prior handle rather than conservatively keeping it forever.

I also rechecked the increment against the surrounding alias/container model. Ordinary `const alias = controls` continues to share the exact resolved object value, conditional/ambiguous container targets remain conservative, and the change does not alter runtime or spec semantics. All current review threads are resolved.

Exact-head trusted repository CI run `33807115705` is green, including `test` (`Public documentation gate and repository tests`), `type-check`, `lockstep-check`, `release-scan`, and package/release lanes; Poppy preview run `33807115718` is green; DCO reports all commits signed off. I found no remaining blocking mismatch on this head.

PR #604, opened 2026-09-03
test(docs): assert the Tooltip Group through the runtime host

### LINKED ISSUE #603: bug(tests): the Brutalist Tooltip Group browser case asserts a native select the previewer never renders
The browser case added by #502 fails on `main` itself, so it blocks every open PR. Confirmed by running it on `main@28a94f7c` with no other changes.

```
FAIL apps/www/src/content/docs/zh-cn/demo-brutalist-controls.browser.test.ts
  > composes the transparent Tooltip Group across Web Components, React, and Vue
AssertionError: expected [] to deeply equal [ 'wc', 'react', 'vue' ]
```

## Cause

The first assertion reads the runtime list from a native `<select>`:

```ts
await previewer
  .locator('select.adapter-select option')
  .evaluateAll((options) => options.map((option) => (option as HTMLOptionElement).value))
```

The previewer renders no native `<select>` at all. Probing `/en/ui-libraries/brutalist/components/tooltip/` in Chromium:

```json
{
  "hasPreviewer": true,
  "selectAdapterClass": 0,
  "anySelect": [],
  "adapterSelectRoot": 1,
  "optionValues": [],
  "puiRoots": 14
}
```

The adapter selector is a Proto UI Select under `[data-adapter-select-root]` — which is what every other case in this suite uses, via `runtimeSelectTrigger` and `choosePreviewRuntime` in `browser-harness.ts`. `select.adapter-select` matches nothing, so `evaluateAll` returns `[]` before any runtime is exercised.

The same probe reports **14** `[data-pui-root]` nodes on the page, while the case asserts **7** after `selectRuntime`. Worth checking alongside the selector fix; I did not get past the first assertion to see whether the 7 is right for one runtime.

## Also

`/en/ui-libraries/brutalist/components/tooltip/` is not in `READY_ROUTES` in `scripts/test/run-runtime-tests.mjs`, so the shared dev server never warms it. That is the cold-Astro-route problem the comment there already describes, and it will make this case slow or flaky once the selector is fixed.

## Not caused by an open PR

I checked this from my side before reporting: on #601 I generated the documentation token closure with and without that PR's extractor changes and the two are byte-identical, and the case fails the same way on `main@28a94f7c` detached with nothing of mine applied.

HyacinthHaru:
Fixes #603 under the bounded repair decided there. Test and harness only — no Tooltip, Previewer, Adapter, spec, or runtime change.

## The three changes

**1. Removed the option-inventory assertion.** The case opened with

```ts
await previewer.locator('select.adapter-select option').evaluateAll(…)
```

There is no native `<select>` in the previewer: the adapter selector is `wc-shadcn-select-root[data-adapter-select-root]`, `.adapter-select` is a class on its `wc-shadcn-select-trigger`, and the items live in a popover that only exists while it is open. So the expression matched nothing, returned `[]`, and the case aborted before exercising a single runtime.

Per the decision I did not teach it to open the selector. Choosing each runtime is the stronger evidence and the loop already does it — `selectRuntime` waits for `[data-adapter-select-root].dataset.value` to equal the runtime before accepting the host, so a runtime that cannot be chosen fails there rather than in an inventory read.

**2. Scoped the demo-root count to the host.** `previewer.locator('[data-pui-root]')` spans the previewer chrome, which is itself Proto UI, so the count included selector roots. It is now `.host [data-pui-root]`, matching what `selectRuntime` counts, and the same locator carries the trigger lookups below it. The `7` expectation holds under that scope — verified now that the first assertion no longer aborts.

**3. Added the tooltip route to `READY_ROUTES`.** `run-runtime-tests.mjs` warms every route the browser suites wait on, and this one was missing since #502.

## Evidence

The case passes across Web Components, React, and Vue, and the whole Brutalist browser suite is green:

```
✓ composes the transparent Tooltip Group across Web Components, React, and Vue  5310ms
  Tests  10 passed (10)
```

`check:types` clean. Found while diagnosing a red `test` job on #601, which this failure was blocking along with every other open PR.

### DIFF
--- a/apps/www/src/content/docs/zh-cn/demo-brutalist-controls.browser.test.ts
+++ b/apps/www/src/content/docs/zh-cn/demo-brutalist-controls.browser.test.ts
@@ -753,15 +753,16 @@ describe.sequential('Brutalist control documentation browser regressions', () =>
     };
 
     try {
-      expect(
-        await previewer
-          .locator('select.adapter-select option')
-          .evaluateAll((options) => options.map((option) => (option as HTMLOptionElement).value))
-      ).toEqual([...RUNTIMES]);
+      // Choosing each declared runtime is the evidence: `selectRuntime` waits
+      // for `[data-adapter-select-root].dataset.value` to equal it before the
+      // host is accepted. Reading an option inventory would retest the
+      // previewer's own Select, whose items exist only while it is open.
       for (const runtime of RUNTIMES) {
         await applyColorScheme(page, 'light');
         await selectRuntime(page, previewer, runtime, '[data-pui-root]', 7);
-        const roots = previewer.locator('[data-pui-root]');
+        // Scoped to the rendered host: the previewer chrome is Proto UI too, so
+        // a previewer-wide count is not evidence about this demo.
+        const roots = previewer.locator('.host [data-pui-root]');
         expect(await roots.count(), runtime).toBe(7);
         expect(await roots.nth(0).getAttribute('data-pui-root'), runtime).toBe('');
         const firstTrigger = roots.filter({ hasText: 'Hover or focus for details' }).last();
--- a/scripts/test/run-runtime-tests.mjs
+++ b/scripts/test/run-runtime-tests.mjs
@@ -24,6 +24,7 @@ const READY_ROUTES = [
   '/en/ui-libraries/brutalist/components/button/',
   '/en/ui-libraries/brutalist/components/switch/',
   '/en/ui-libraries/brutalist/components/tabs/',
+  '/en/ui-libraries/brutalist/components/tooltip/',
   '/en/ui-libraries/brutalist/components/select/',
   '/en/ui-libraries/shadcn/checkbox/',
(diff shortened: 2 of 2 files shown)


cyjin-yl (approved):
Reviewed exact head `80cbc3dcb64686d417ffc7f085201a3ee5210f13` against the bounded repair accepted in #603. The patch removes the invalid native-select inventory assertion without adding a second Previewer-Select test path, scopes the demo-root evidence to `.host`, and adds the Tooltip route to `READY_ROUTES`. The affected browser case therefore exercises runtime selection through the existing harness rather than previewer chrome, and the change stays confined to browser test/harness warmup with no Tooltip/Adapter/spec semantic change.

Trusted repository CI for this exact head is green, including the real `test` job (`Public documentation gate and repository tests`), `type-check`, `lockstep-check`, and `release-scan`; DCO is also green. The only commit-status failure is the fork Vercel authorization status, which is not a repository-code failure. There are no review threads or prior blocking reviews on this PR. No blocking finding remains.

PR #605, opened 2026-09-03
test(brutalist): measure Dialog overlay geometry in a real browser (#539)

HyacinthHaru:
Closes the last #539 slice with no carrier. Badge / Card / Hover Card / Separator / Skeleton / Toggle stay on #571; this touches only Dialog.

## What was missing

`T-BRUTALIST-DIALOG-0001`'s ten cases are all consumed by one happy-dom module test that asserts token strings with `styleContains`. Two criteria say more than a token string can:

- `P-BRUTALIST-DIALOG-MASK-VISUAL-GRAMMAR` — a flat scrim over the viewport, no blur.
- `P-BRUTALIST-DIALOG-CONTENT-VISUAL-GRAMMAR` — a centred, square, hard-shadowed panel.

happy-dom lays nothing out, so "covers the viewport", "centred" and "above the mask" hold nowhere in the repository. The existing browser coverage for Dialog is `demo-matrix.browser.test.ts`, which exercises **Base** Dialog's `aria-haspopup`/`aria-controls` and focus, never the Brutalist projection and never any geometry.

## What this adds

One case, `T-BRUTALIST-DIALOG-0001-CASE-OVERLAY-GEOMETRY-BROWSER`, in a new suite on the existing `/en/ui-libraries/brutalist/components/dialog/` route and the existing `demo-brutalist-dialog.demo.ts` carrier — no new demo. It runs in Web Components, React, Vue and Vue 2, and per runtime asserts the mask rect equals the viewport, the panel centre equals the viewport centre, the panel outranks the mask and is the element under its own centre, `backdrop-filter` resolves to `none`, and the composed `box-shadow` carries `rgb(0, 0, 0) 3px 3px 0px 0px`. It then changes the viewport and re-measures, so a projection that positioned the panel once at mount cannot pass on the first measurement.

Each of those was checked against a mutation of the **source**, not of the test:

| mutation | fails on |
| --- | --- |
| drop `-translate-x-1/2 -translate-y-1/2` from Content | `wc/initial/centre-x` |
| drop `inset-0` from Mask | `wc/mask-y` |
| add `backdrop-blur-sm` to Mask | `wc/mask-no-blur` |
| `shadow-[3px_3px_0_0_#000]` → `shadow-none` in `BRUTALIST_PANEL_TOKENS` | `wc/hard-shadow` |

The square-corner assertion is deliberately not in that table: a radius the token closure never generated also computes to `0px`, so it states the rendered result without separating those two causes. The comment says so.

## One thing found on the way

Overlays are selected by `data-is-present`, not by body membership, because **Vue 2 leaves a closed Dialog's portalled mask and content in `document.body`, and they outlive the app that mounted them** — they are still there after switching the previewer to another runtime. Web Components, React and Vue 3 all remove them on close. The residue is inert (`display: none`, `visibility: hidden`, `0×0`), so nothing renders wrong, but a body-child count reads a previous runtime's leftovers as the current runtime's overlay. Filed as #606; I did not touch the adapter here, since where a non-present overlay parks is a teardown decision rather than something this test should settle.

## Checks

`packages/spec` 44 passed — including `catalog-evidence-integrity`, which caught the new file before it was tracked. `check:types` clean, `check:prototype-catalog` OK. The suite passes all four runtimes in ~6s.

### DIFF
--- a/apps/www/src/content/docs/zh-cn/demo-base-controls.browser.test.ts
+++ b/apps/www/src/content/docs/zh-cn/demo-base-controls.browser.test.ts
@@ -381,6 +381,23 @@ async function setFittingContent(page: Page, rows: number): Promise<void> {
  * with or without the focus slice, so a missing tab stop fails on the assertion
  * that names it instead of on a readiness timeout that does not.
  */
+/**
+ * A wheel notch is delivered asynchronously and the scroll may be smoothed, so
+ * a fixed pause races the browser under load. Wait for the offset instead.
+ */
+async function waitForScrolled(page: Page, index: number): Promise<void> {
+  await page.waitForFunction(
+    ({ selector, index: at }) => {
+      const viewport = document.querySelectorAll<HTMLElement>(`[data-previewer-id] ${selector}`)[
+        at
+      ];
+      return Boolean(viewport && viewport.scrollTop > 0);
+    },
+    { selector: VIEWPORT_SELECTOR, index },
+    { timeout: 10_000 }
+  );
+}
+
 async function waitForCanScroll(page: Page, index: number, expected: boolean): Promise<void> {
   await page.waitForFunction(
     ({ selector, index: at, expected: want }) => {
@@ -606,7 +623,7 @@ describe.sequential('Base control documentation browser regressions', () => {
 
         await viewport.hover();
         await page.mouse.wheel(0, 120);
-        await page.waitForTimeout(200);
+        await waitForScrolled(page, SCROLLING);
         expect(
           (await viewportFacts(page))[SCROLLING].scrollTop,
           `${runtime}/wheel`
--- a/apps/www/src/content/docs/zh-cn/demo-brutalist-dialog.browser.test.ts
+++ b/apps/www/src/content/docs/zh-cn/demo-brutalist-dialog.browser.test.ts
@@ -1,0 +1,219 @@
+// @vitest-environment node
+
+import type { Browser, Page } from 'playwright-core';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import {
+  RUNTIMES,
+  launchBrowser,
+  openRoute,
+  selectRuntime,
+  startServer,
+  stopServer,
+} from './browser-harness';
+
+const DIALOG_ROUTE = '/en/ui-libraries/brutalist/components/dialog/';
+
+const VIEWPORT = { width: 1280, height: 900 } as const;
+
+/** A second viewport, so a panel centred once cannot pass by standing still. */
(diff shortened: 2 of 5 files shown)


cyjin-yl (commented):
Reviewed exact head `dcdec466f94f0f0115fecc2472721370b76d2487`, including the new browser carrier, runtime-test plan/warmup registration, and `T-BRUTALIST-DIALOG-0001` evidence binding. I do not have a code/spec blocker from this diff: the browser test measures the current Mask/Content contract rather than just token strings, scopes overlay selection to present roots (so stale/non-present nodes cannot contaminate a runtime), re-measures after viewport resize, and covers WC/React/Vue/Vue2. The focused Dialog browser case itself passes on the exact-head repository run, as do `type-check`, `lockstep-check`, `release-scan`, docs preview, and DCO.

I am not approving this head yet because the trusted repository CI run `33747017641` is red. The failing repository test is outside this PR's changed files: `apps/www/src/content/docs/zh-cn/demo-base-controls.browser.test.ts` fails `keeps keyboard and wheel scrolling on the Viewport, in all runtimes` at the WC wheel assertion (`expected 0 to be greater than 0`, line 613). In the same run, `demo-brutalist-dialog.browser.test.ts` passes for all four runtimes. That evidence is insufficient to request a Dialog/test change here, but the merge gate still requires a successful exact-head repository CI. Please obtain a green trusted CI on this exact head (or, if the WC wheel failure is reproducible, fix/land that independent repository regression first and rerun this PR on the updated base).

The Vercel fork-authorization status is not being treated as a code failure. Also, #606 is correctly separate: this test's `data-is-present` selection remains robust whether Vue 2 retains a merely closed overlay or later fixes app-teardown ownership.

cyjin-yl (approved):
Reviewed current exact head `625a545d62da437b9e26dbae74d84cc58127c637`, including the increment since my prior COMMENT. The Dialog browser evidence remains bounded to the existing visual-grammar contract and still measures the real rendered geometry across WC/React/Vue/Vue2 without depending on the separate Vue 2 teardown decision in #606. The added `demo-base-controls.browser.test.ts` change is a narrow synchronization repair for the exact WC wheel failure that blocked the previous head: it replaces a fixed 200 ms sleep with a bounded wait for the asserted `scrollTop > 0` condition and changes no product/runtime semantics.

Current-head trusted CI `33748507742` is green, including `test`, `type-check`, `lockstep-check`, and `release-scan`; the docs preview is green and DCO reports all commits signed off. There are no unresolved review threads. The only failing external status is the fork Vercel authorization link, which is not treated as a repository-code failure. I found no remaining blocker on this exact head.

PR #608, opened 2026-09-03
fix(vue2): return a portalled overlay host when the mount unwinds (#606)

### LINKED ISSUE #606: bug(vue2): a closed Dialog's portalled nodes stay in document.body and outlive the app
## Summary

When a Brutalist Dialog closes under the Vue 2 adapter, its portalled Mask and Content stay in `document.body`. They are still there after the Vue 2 app that mounted them is unmounted. Web Components, React and Vue 3 all remove them on close.

## Evidence

Measured on `main@fe7d9714`, Chromium, `/en/ui-libraries/brutalist/components/dialog/`, switching the previewer runtime and listing `document.body`'s children that carry `data-pui-root`:

```
react after switch: []
react open:   ["div role=null  state=entered present=true  display=block vis=visible rect=1280x900",
               "div role=dialog state=entered present=true display=grid  vis=visible rect=512x188"]
react closed: []

vue after switch: []
vue open:     [... same shape ...]
vue closed:   []

vue2 after switch: []
vue2 open:    [... same shape ...]
vue2 closed:  ["div role=null   state=closed present=false display=none vis=hidden rect=0x0",
               "div role=dialog state=closed present=false display=none vis=hidden rect=0x0"]

wc after switch: ["div role=null   state=closed present=false display=none vis=hidden rect=0x0",
                  "div role=dialog state=closed present=false display=none vis=hidden rect=0x0"]
```

The last line is the point: those two `div`s are the Vue 2 leftovers, still in the body after the previewer moved on to Web Components and the Vue 2 app was torn down.

## Impact

Nothing renders wrong — the residue is `display: none`, `visibility: hidden`, `0×0`, `data-is-present` absent. Two consequences are real:

1. Every open/close cycle in a long-lived Vue 2 page leaves nodes behind that no live component owns.
2. Any consumer that reasons about overlays by walking `document.body` sees a stale pair. The Dialog browser suite in #605 selects by `data-is-present` for exactly this reason; a body-child count read the previous runtime's residue as the current runtime's overlay.

## Not proposed here

Where a non-present overlay parks — removed, or kept detached in place — looks like a deliberate difference between the adapters rather than an obvious bug, and Web Components parks the closed pair inside the demo host rather than removing it outright. What is asked here is only that teardown of the app removes what that app portalled.

## Classification
- Area: adapters (vue2), prototypes (dialog)
- Priority: F3 – Moderate

HyacinthHaru:
Closes #606, within the bounds you set: Vue 2 adapter only, no change to Dialog/Presence semantics, the Overlay contract, or where a merely closed overlay parks.

## The repair

`createVue2OverlayGlobalMount().unmount()` removed the anchor and cleared the logical projection but never moved the host back, so the body-owned node had no owner left to remove it. It now takes the reversible route you preferred: while the anchor still has a parent, the host is inserted back **before the anchor** — its original position, not merely its original container — and only then is the anchor removed, so Vue's ordinary teardown removes the restored node. If the anchor's parent is gone there is nowhere to return it to, and the host is detached rather than left behind.

Only a host this mount actually moved is touched: `mount()` records an anchor only when it moves something, so the early-return cases (no parent, already in `body`) own nothing and `unmount()` leaves them alone. That also makes repeated unmounts inert.

## Evidence

You were right that the existing test masked the gap — it asserted the original container was empty, which the leak satisfied trivially, and then removed the host by hand. That assertion is now the opposite: the container holds the host again.

Three tests, all three of which fail against the old `unmount` and pass against the new one:

- **Anchor valid.** Mount moves the host to `body` and leaves one anchor; unmount returns it *between the same siblings* (`[before, childRoot, after]`), removes the anchor, and leaves nothing of it in `body`. A second unmount moves nothing.
- **Anchor orphaned.** The original container is removed while the host is portalled; unmount detaches the host rather than leaving an adapter-owned body node, and leaves no anchor.
- **App teardown.** A real adapted Dialog opens, both the Mask and the Content are asserted to be portalled out of the app's subtree, then `vm.$destroy()`; every node that was portalled is asserted disconnected and no portal anchor remains. Nothing closes the dialog first, deliberately — a close-only assertion would not prove teardown reclaimed the DOM, and this does not decide where a non-present overlay parks.

`:scope > …` is not supported by the test DOM, so the body children are filtered rather than selected; the comment says so.

`packages` 1642 passed, `check:types` clean.

### DIFF
--- a/packages/adapters/vue2/src/runtime/modules.ts
+++ b/packages/adapters/vue2/src/runtime/modules.ts
@@ -124,8 +124,18 @@ export function createVue2OverlayGlobalMount(
       anchors.set(hostEl, anchor);
     },
     unmount(hostEl: HTMLElement) {
-      anchors.get(hostEl)?.remove();
+      const anchor = anchors.get(hostEl);
       anchors.delete(hostEl);
+      // Only a host this mount moved is ours to put back. Returning it to the
+      // anchor hands it to the position Vue owns, so ordinary teardown removes
+      // it; if the anchor's parent is gone there is nowhere to return it to and
+      // detaching is the only way not to leave an adapter-owned body node.
+      if (anchor) {
+        const parent = anchor.parentNode;
+        if (parent) parent.insertBefore(hostEl, anchor);
+        else hostEl.remove();
+        anchor.remove();
+      }
       // The retained Proto instance stays in its logical tree while its Vue2
       // view epoch is torn down and will receive a fresh projection on remount.
       clearProtoParentProjection(hostEl);
--- a/packages/adapters/vue2/test/overlay-portal-ownership.test.ts
+++ b/packages/adapters/vue2/test/overlay-portal-ownership.test.ts
@@ -1,6 +1,13 @@
 import { describe, expect, it } from 'vitest';
 import { definePrototype } from '@proto.ui/core';
+import {
+  dialogContent,
+  dialogMask,
+  dialogRoot,
+  dialogTrigger,
+} from '../../../prototypes/base/src/dialog';
 
+import { createVue2Adapter } from '../src/adapt';
 import { createVue2OverlayGlobalMount } from '../src/runtime/modules';
 import {
   bindLogicalParent,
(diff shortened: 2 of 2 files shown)


cyjin-yl (approved):
Reviewed exact head `d4789f2cc373fd0e23db2d072221dad48557a283` against the bounded #606 maintainer ruling. The Vue 2 global-mount teardown now restores only hosts this mount actually moved, using the saved anchor to preserve sibling order; if the anchor has lost its parent it detaches the host instead of leaving a body orphan, and repeated unmount is inert. The focused tests cover normal restore/order, invalid-anchor cleanup, and a real Vue 2 Dialog app teardown proving portalled `data-pui-root` nodes and anchors are disconnected. This stays within the adapter lifecycle boundary and does not change Dialog/Presence or close-only semantics.

Exact-head trusted repository CI is green, including `test`, `type-check`, `lockstep-check`, `release-scan` and release/package lanes; DCO is green. The only failing external status is Vercel fork deployment authorization, which is not a repository-CI failure. There are no unresolved review threads or other blocking findings.

PR #610, opened 2026-09-03
fix(cli): keep the container a conditional replacement may not install (#609)

### LINKED ISSUE #609: bug(cli): conditional container replacement loses the prior state identity
## Summary

Two shapes where the static token extractor and the coverage gate agree on the wrong state identity while the gate stays green — the failure class #601 exists to eliminate. Both are on `main` as of `30ca6ef6`. Both are latent: no shipped prototype uses either shape, and the repository sweep is unchanged at 118 usages / 1 exposed local / 0 unresolved.

These were found by the reviewer's independent review of `55d0fb5d` during #601 but never reached that PR — its submission fails closed when the PR state changes, and the findings were discarded. #601 was approved and merged on a later head without them.

## R-01 — a conditional whole-container reassignment drops the prior members

```ts
const first = def.state.bool('firstFlag', false);
const second = def.state.bool('secondFlag', false);
let controls = { ready: first };
if (enabled) controls = { ready: second };
def.expose.state('visible', controls.ready);
def.rule({ when: (w) => w.state(controls.ready).eq(true), intent: use });
```

Emits `data-[second-flag]:…` only. When `enabled` is false the runtime exposes and lowers `first`, so the optimized rule has no CSS. The gate reports nothing.

The assignment is correctly treated as uncertain, but only the binding's own semantic is carried over — the replaced container's member table is not, so the read path sees the new members alone.

## R-02 — a conditional-expression replacement is recorded by neither prepass

```ts
let controls = { ready: first };
controls = enabled ? { ready: second } : { ready: third };
def.expose.state('visible', controls.ready);
def.rule({ when: (w) => w.state(controls.ready).eq(true), intent: use });
```

Emits **no variant at all** — only the bare token — and reports nothing unresolved. `aliasTargets` fans out over conditional branches that are identifiers, and `recordObjectLiteral` runs only when the right-hand side is directly an object literal, so a conditional of two literals matches neither. The runtime lowers whichever branch it took; that CSS does not exist.

## Acceptance

Same bar as #601: the emitted selector set must be a superset of what the Web optimizer can emit, spelled exactly per candidate, and the gate must not certify a shape the extractor cannot lower. Extractor and coverage regressions for both shapes, in both analyzers, each mutation-checked.

## Classification
- Area: cli (prototype-style-tokens, lowered-hook-coverage)
- Priority: F2 — latent, but silent-wrong with a green gate

HyacinthHaru:
Closes #609. Both shapes reproduced on `main@30ca6ef6` before anything was changed, and both had the signature #601 exists to eliminate: extractor and gate agreeing on a state identity the runtime does not use, with nothing reported.

## R-01 — a conditional replacement dropped what it replaced

`if (enabled) controls = { ready: second }` emitted `data-[second-flag]:…` alone. The assignment was already treated as uncertain, but only the binding's own *semantic* carried over — the replaced container's member table did not, so the read path saw the new members only. A container replaced under uncertainty now keeps the one it replaced among the values a member read consults, in both analyzers.

## R-02 — a conditional of two literals was recorded by neither prepass

`controls = enabled ? { ready: second } : { ready: third }` emitted **no variant at all** and reported nothing unresolved. `aliasTargets` fans out over conditional branches that are *identifiers*, and `recordObjectLiteral` ran only when the right-hand side was *directly* an object literal, so a conditional of two literals fell between them. Both prepasses now take every object literal an initializer may yield, each recorded conditionally.

Note what this deliberately does not add: `first` gets no variant here, because that assignment always runs and the container it replaced is genuinely gone. Only R-01's skippable replacement retains the prior container. The two regressions assert opposite things on that point, so the fix cannot degrade into "always keep everything".

## Verification

Four new regressions across both analyzers, each mutation-checked: dropping the replaced container turns two red, and restricting the prepasses to a direct literal turns two red.

Repository sweep unchanged at **118 usages, 1 exposed local, zero unresolved** — both shapes are latent, as #609 records. `packages` 1644 passed, `check:types` clean, preset manifests still 246 / 236.

## Provenance

These are the reviewer's R-01 and R-02 from its independent review of `55d0fb5d` during #601. That submission failed closed on PR-state drift and the findings were discarded, so #601 was approved and merged on a later head without them; the review noted it had not re-verified whether the later head folded them in. It had not.

### DIFF
--- a/packages/cli/src/services/lowered-hook-coverage.ts
+++ b/packages/cli/src/services/lowered-hook-coverage.ts
@@ -91,6 +91,8 @@ type Binding =
       initializer: ts.Expression;
       /** Members written after the declaration, latest candidates first. */
       members?: Map<string, ts.Expression[]>;
+      /** Containers the name may still hold because a write may be skipped. */
+      alternates?: ts.Expression[];
     }
   /** A named import; only a relative one is something the extractor follows. */
   | { kind: 'tokenImport'; specifier: string; imported: string; from?: string }
@@ -394,6 +396,19 @@ export function scanRuleStateReads(
   type MemberEdge = { target: string; at: number; conditional: boolean };
   const objectMembers = new Map<ts.Node, Map<string, Map<string, MemberEdge[]>>>();
 
+  /**
+   * Every object literal an initializer may yield. A conditional selects one at
+   * runtime, so both are recorded for the same reason `aliasTargets` fans out.
+   */
+  const objectLiteralTargets = (node: ts.Node): ts.ObjectLiteralExpression[] => {
+    const value = unwrap(node);
+    if (ts.isObjectLiteralExpression(value)) return [value];
+    if (ts.isConditionalExpression(value)) {
+      return [...objectLiteralTargets(value.whenTrue), ...objectLiteralTargets(value.whenFalse)];
+    }
+    return [];
+  };
+
   /** Every member an object literal names, at the position it was written. */
   const recordObjectLiteral = (
     owner: ts.Node,
@@ -627,10 +642,16 @@ export function scanRuleStateReads(
       names.add(node.name.text);
       declaredIn.set(owner, names);
       if (node.initializer) {
-        const literal = unwrap(node.initializer);
         const conditional = isConditionallyReached(node);
-        if (ts.isObjectLiteralExpression(literal)) {
-          recordObjectLiteral(owner, node.name.text, literal, node.getStart(), conditional);
+        const literals = objectLiteralTargets(node.initializer);
+        for (const literal of literals) {
+          recordObjectLiteral(
+            owner,
+            node.name.text,
+            literal,
+            node.getStart(),
+            conditional || literals.length > 1
+          );
         }
         for (const target of aliasTargets(node.initializer)) {
           aliasEdges.push({
@@ -721,9 +742,17 @@ export function scanRuleStateReads(
         nextChain[0] ??
         source;
       const conditional = isConditionallyReached(node);
-      const replacement = unwrap(node.right);
-      if (ts.isObjectLiteralExpression(replacement)) {
-        recordObjectLiteral(assignOwner, left, replacement, node.getStart(), conditional);
+      const replacements = objectLiteralTargets(node.right);
+      for (const replacement of replacements) {
+        recordObjectLiteral(
+          assignOwner,
+          left,
+          replacement,
+          node.getStart(),
+          // Either literal may be the one the runtime took, and the name may
+          // still hold the container it replaced.
+          conditional || replacements.length > 1
+        );
       }
       for (const target of aliasTargets(node.right)) {
         aliasEdges.push({
@@ -1163,7 +1192,11 @@ export function scanRuleStateReads(
         // A write through the container replaces what the declaration named.
         const written = binding.members?.get(key);
         if (written && written.length > 0) return written;
-        return containerMembers(binding.initializer, key, scope, new Set([...seen, value.text]));
+        const next = new Set([...seen, value.text]);
+        // A container that replaced another conditionally still answers for it.
+        return [binding.initializer, ...(binding.alternates ?? [])].flatMap((held) =>
+          containerMembers(held, key, scope, next)
+        );
       }
     }
     return [];
@@ -1807,7 +1840,21 @@ export function scanRuleStateReads(
     ) {
       const name = node.left.text;
       const owner = scopeBinding(current, name) ?? current;
+      const replaced = owner.bindings.get(name);
       declareConditionally(owner, name, node.left, current, node.right);
+      const next = owner.bindings.get(name);
+      const uncertain =
+        isConditionallyReached(node) || isDeferredWrite(node, enclosingFunction(owner.node));
+      if (uncertain && replaced?.kind === 'token' && next?.kind === 'token') {
+        declare(owner, name, {
+          ...next,
+          alternates: [
+            ...(next.alternates ?? []),
+            replaced.initializer,
+            ...(replaced.alternates ?? []),
+          ],
+        });
+      }
     }
 
     if (
--- a/packages/cli/src/services/lowered-hook-coverage.ts
+++ b/packages/cli/src/services/lowered-hook-coverage.ts
@@ -996,12 +996,23 @@ export function scanRuleStateReads(
           declaredAs: binding.declaredAs,
           exposedAs: binding.exposedAs,
         });
+        // A branch may itself be a conditional, so its own alternatives are
+        // lifted here; `head.alternatives` would otherwise be overwritten by
+        // this key and `bare` would strip the rest's.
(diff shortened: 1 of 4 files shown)


cyjin-yl (changes requested):
Blocking: nested conditional container replacements still lose a runtime-reachable member identity in the production extractor.

On this exact head, `packages/cli/src/services/prototype-style-tokens.ts` represents a conditional container as `{ ...branches[0], containers }`, but `containers` is built only from the two direct branch values whose own `semanticMap` is present. `readMemberSemantics()` then consults only each direct container's `semanticMap`; it does not recursively flatten/consult a container value's own `containers`.

That makes nesting unsound. For example:

```ts
const first = def.state.bool('firstFlag', false);
const second = def.state.bool('secondFlag', false);
const third = def.state.bool('thirdFlag', false);
let controls = a
  ? (b ? { ready: first } : { ready: second })
  : { ready: third };
def.expose.state('visible', controls.ready);
def.rule({
  when: (w) => w.state(controls.ready).eq(true),
  intent: (i) => i.feedback.style.use(tw('bg-accent')),
});
```

The inner conditional resolves to a composite based on its first branch with `containers: [firstObject, secondObject]`. When that composite becomes the outer conditional's first branch, the outer `containers` becomes `[innerComposite, thirdObject]`. A later member read sees `innerComposite.semanticMap` (the `first` branch) and `thirdObject.semanticMap`, but never descends into `innerComposite.containers`, so the `second` identity can disappear even though the runtime can choose it. The Web optimizer may therefore lower to `data-[second-flag]` with no emitted matching variant.

Fix criterion: recursively flatten container alternatives when constructing conditional bindings, or recursively traverse them when resolving a member, so every runtime-reachable leaf container contributes its member identity. Please add paired regressions for a nested conditional in the extractor and coverage scanner; the extractor must emit first/second/third, and the gate must not certify the shape unless all runtime-reachable leaves are represented. Preserve the current distinction that an unconditional whole-container replacement does not retain the container it definitely replaced.

The repository CI and DCO are green on this head; the fork Vercel authorization status is not part of this finding.

cyjin-yl (changes requested):
One blocking fail-closed gap remains on exact head `f1bd9ec5` in the conditional-container path this PR adds.

The new candidate counting only counts branches that are recognized as an object literal or an identifier alias:

- `objectLiteralTargets(...)` recurses through a conditional but returns `[]` for every other expression;
- `aliasTargets(...)` likewise returns `[]` for every branch that is not an identifier/conditional;
- assignment handling then uses `valueCount = replacements.length + aliasTargets(node.right).length` to decide whether the recorded literal/alias edge is conditional.

That makes a mixed *recognized + opaque* conditional unsound. For example:

```ts
const second = def.state.bool('secondFlag', false);
const third = def.state.bool('thirdFlag', false);
const makeControls = () => ({ ready: third });
let controls = { ready: second };
controls = enabled ? { ready: second } : makeControls();
def.expose.state('visible', controls.ready);
def.rule({
  when: (w) => w.state(controls.ready).eq(true),
  intent: (i) => i.feedback.style.use(tw('bg-accent')),
});
```

At runtime the false branch can leave `controls.ready === third`, so the Web optimizer can lower the rule to the `third-flag` attribute. Statically, however, `objectLiteralTargets(node.right)` sees only the true-branch literal, `aliasTargets(node.right)` sees no branch at all, so `valueCount === 1`; the literal is recorded as if it were the only possible value. The production `resolveBinding` also contributes only the recognized literal container. The coverage scanner has the same blind spot: `containerMembers()` recurses into both conditional branches but returns `[]` for the call expression, so it can report only `second-flag` with no unresolved subject. Extractor and gate can therefore agree on an incomplete set and stay green.

This does not require statically evaluating arbitrary calls. Fix criterion: when a conditional/container expression has any runtime branch that is not proven to one of the supported container forms, do not treat the recognized branches as exhaustive. Either conservatively retain all provable candidates *and* make the coverage gate fail closed for the opaque branch, or add a sound resolution path for that form. Please add paired extractor/gate regressions for a conditional with one literal (or identifier) branch and one opaque call-expression branch; the gate must not certify the shape unless every runtime-reachable container identity is represented.

The previous nested-conditional, F-02, and F-03 findings appear fixed on this head, the branch is rebased onto current `main@64266b56`, trusted repository CI and DCO are green, and there are no review threads. The fork Vercel authorization failure is not part of this finding.

cyjin-yl (changes requested):
One blocking gate-oracle gap remains on exact head `fd9c68e773adcfb273f7ae8be3562525973f9913`.

The previous mixed recognized+opaque conditional finding is fixed on this head: the extractor now counts all conditional leaves, and the coverage scanner fails closed when a container is only partially recognized. Trusted repository CI and DCO are green, and there are no review threads.

The remaining problem is in `packages/cli/src/services/lowered-hook-coverage.ts`, in the local-state conditional merge inside `declareValue`. For a nested conditional such as:

```ts
const first = def.state.bool('firstFlag', false);
const second = def.state.bool('secondFlag', false);
const third = def.state.bool('thirdFlag', false);
const current = a ? (b ? first : second) : third;
def.expose.state('visible', current);
def.rule({
  when: (w) => w.state(current).eq(true),
  intent: (i) => i.feedback.style.use(tw('bg-accent')),
});
```

the inner branch can resolve to a local-state binding whose `declaredAs` is `first` and whose `alternatives` contains `second`. The outer merge then spreads `head` but immediately replaces its `alternatives` with `rest.map(bare)`, so the nested `second` alternative is discarded. If the nesting is on the other side, `bare()` similarly strips the nested alternatives from the `rest` binding. The scanner can therefore certify only two of the three runtime-reachable state identities with `unresolved` empty.

The production extractor is already recursive here via `bindingSemantics`, so this is currently gate under-certification rather than a manifest selector miss. It is still blocking because the coverage gate must not certify an incomplete runtime candidate set: otherwise a future extractor regression on this shape can pass the oracle silently.

Fix criterion: when merging conditional `LocalStateBinding`s, recursively lift alternatives from both the head and every remaining branch, then dedupe by `declaredAs`. Add scanner regressions for both nesting orientations (`a ? (b ? first : second) : third` and `a ? first : (b ? second : third)`) that require all three exposed local attributes and an empty `unresolved` result. Preserve the existing ordinary two-branch behavior.

cyjin-yl (approved):
Re-reviewed exact head `b6a6ae443e6b7ff786285608374fe8c317a99c6b`. The remaining gate-oracle blocker from the prior review is fixed: nested local-state conditional merges now lift `head.alternatives` plus each remaining branch and its alternatives, dedupe by `declaredAs`, and the regression covers both requested nesting orientations plus the ordinary two-branch boundary. The earlier conditional-container findings remain addressed in the current diff. Trusted repository CI and DCO are green on this head, there are no unresolved review threads, the branch is current with `main@64266b56`, and the only failing external status is the fork Vercel deployment authorization. I found no remaining blocking correctness, scope, contract, test, lifecycle, or integration issue on this exact head.

PR #627, opened 2026-09-07
feat(context): catalog M/HC/A slice and verify Vue 2 conformance

guangliang2019:
## Summary

Catalog Context as a coherent Module/Host Capability/Adapter slice and close concrete Vue 2 follow-through gaps. Add draft `M-CONTEXT-0001`, identity/ancestry HC entities, four official Adapter support relations, and `T-CONTEXT-0003` with real-framework tests.

The Vue 2 keep-alive audit exposed a shared Adapter reader regression: replacement expose snapshots recreated top-level method wrappers across view detachment. Preserve wrapper identity while retaining the current root receiver, distinct nested receivers, callback scope, and terminal invalidation.

## Related context

- Issue: maintainer-directed Context cataloging and Vue 2 follow-through audit; no automatic issue closure.
- Applicable entities: draft `C-CONTEXT-0001..0012`, `M-CONTEXT-0001`, `HC-CONTEXT-IDENTITY-0001`, `HC-CONTEXT-ANCESTRY-0001`; existing official React, Vue 3, Vue 2 and Web Component A profiles.
- Criteria / evidence: `C-CONTEXT-0008-C`, `C-CONTEXT-0012-A/B/C`, `C-EXPOSE-0008-B`; `T-CONTEXT-0002/0003`, `T-PROPS-0012`, `T-EVENT-0003`, `T-STATE-0005`, and Expose-family T entities.
- Engineering record: `internal/records/2026-09-07-context-module-host-cap-adapter-audit.zh-CN.md`.

## Scope boundary

- The maintainer explicitly accepted provider self-update without subscription; consumers still require subscription. The exception does not change read, tryRead or tryUpdate authority.
- Clarify old Context unmount wording using the existing repeatable-view versus terminal-instance lifecycle authority. Reconcile legacy readable projections.
- Add Vue 2 evidence for Props normalization/invalidation, default action/global event cleanup, keep-alive State/Expose retention and current outward listeners.
- No lifecycle promotion, new public author API, non-Web conformance, or complete portal/cross-root relocation matrix.

## Source-of-truth alignment

- [x] Checked applicable spec entities before relying on contracts, records, implementation or docs.
- [x] Updated criteria, revisions, executable evidence, relations and affected legacy projections together.
- [x] No empty catalog identities or implicit stable promotion.

## Contribution provenance

### Third-party or upstream sources

No external code or assets were imported. Tests reuse and adapt existing Proto UI fixtures and Adapter test patterns from baseline `f8f98d22` under this repository's MIT license.

### AI assistance

OpenAI Codex assisted with repository tracing, catalog authoring, tests, the shared reader repair and documentation. The current user decided the provider-update boundary and authorized submission after receiving the change and validation summary. No independent human code review is claimed. Source material used was this repository and current user instructions; no external private code was supplied for this change.

## DCO

- [x] The commit includes a matching Signed-off-by trailer.

## Prototype website preview

- [x] A new or updated website page is not applicable: no public Prototype identity, anatomy or visual demo behavior is introduced. This is Context cataloging, Adapter evidence and a shared Expose lifecycle repair.
- Browser/visual preview: not performed; tests mount real framework owners in happy-dom.

## Validation

Node 22.23.2; pnpm 10.32.1 via Corepack.

- `corepack pnpm@10.32.1 exec vitest run packages/modules/context/test packages/runtime/test/contract/context-lifecycle.v1.contract.test.ts packages/adapters/base/test packages/adapters/react/test packages/adapters/vue/test packages/adapters/vue2/test packages/adapters/web-component/test packages/spec/test` — 188 files, 560 tests passed. Spec suites were separately run with the correct `packages/spec` directory below.
- `corepack pnpm@10.32.1 exec vitest run packages/spec` — 22 files, 61 tests passed. Candidate evidence initially used a temporary Git index; repeated after commit with the real index.
- `corepack pnpm@10.32.1 check:types` — passed (Astro: 188 files, zero diagnostic errors/warnings/hints).
- `corepack pnpm@10.32.1 spec:docs:agent` and `corepack pnpm@10.32.1 workspace:generate` — 572 entities; generated local projections remain ignored.
- `corepack pnpm@10.32.1 check:prototype-catalog`, `check:agent-doc`, `check:agent-operations` — passed; agent operations includes 58 passing tests.
- `git diff --check` and commit-time Prettier — passed.
- Full release-gated `pnpm test`, docs production build and real-browser geometry checks were not run; affected Adapter/module/runtime/spec suites and type checks define this validation scope.

## Context identity and capability boundaries

The final implementation reconciles these boundaries across contract, Context port and its Scroll consumer:

- C-CONTEXT-0004 explicitly resolves from the participant itself, with a clarification revision and executable provider-as-self evidence.
- Null alone denotes identity absence. Falsy tokens (0, false, empty string, undefined and NaN) preserve resolution, updates, callback delivery and terminal cleanup. Identity equality matches the provider Map's SameValueZero semantics.
- The actual Context port preserves argument count. Omission validates the owner with getSelfToken; absent or null owner caps fail diagnostically. Explicit undefined remains a consumer identity, and explicit null deliberately returns no scope without requiring owner identity or ancestry.
- Composed Scroll uses null-only absence and matching identity semantics. Real Context/State/Scroll module tests with supplied Anatomy domain/target facts verify host scrollbar bindings, missing or mismatched scopes, and restoration.

Latest diagnostic correction: f2d1a068. Actual-port tests reproduce both owner-identity failures before the fix and verify missing ancestry, explicit null, and explicit valid consumers independently. The existing T-CONTEXT-0003-CASE-AVAILABILITY mapping covers this evidence; no further semantic change was needed.

Latest validation: `corepack pnpm@10.32.1 exec vitest run packages/modules/context/test packages/modules/scroll/test packages/runtime/test packages/adapters/react/test/context.integration.test.ts packages/adapters/vue/test/context.integration.test.ts packages/adapters/vue2/test/context.integration.test.ts packages/adapters/web-component/test/context.integration.test.ts --maxWorkers=2 --minWorkers=1 --testTimeout=20000` — 264 passed, 34 existing todo (49 passing files, 3 skipped). `check:types:workspace` and `git diff --check` passed.

The preceding integration validation also passed 336 tests including Anatomy and spec, plus full check:types and catalog/projection checks; its initial high-concurrency run hit default 5-second spec timeouts and passed after limiting concurrency without changing assertions. Browser geometry and the full release suite were not rerun locally for these follow-ups. Independent re-review remains pending.

### DIFF
--- a/internal/contracts/context/README.md
+++ b/internal/contracts/context/README.md
@@ -1,60 +1,12 @@
 # Context Contract Index
 
-This directory defines **context** contracts for Proto UI.
+This directory is a readable legacy projection. Applicable `spec/**` entities take precedence; the Context catalog remains **draft**, not a stable release guarantee.
 
-Context is the **component-to-component** communication channel that is:
+- [Context module](../../../spec/modules/M-CONTEXT-0001.yaml): author facade, privileged port, instance-owned resources and standard Runtime composition.
+- [Identity capability](../../../spec/host-caps/HC-CONTEXT-IDENTITY-0001.yaml) and [ancestry capability](../../../spec/host-caps/HC-CONTEXT-ANCESTRY-0001.yaml): opaque owner identity and current logical parent facts.
+- [Read authority](../../../spec/contracts/C-CONTEXT-0007.yaml), [update authority](../../../spec/contracts/C-CONTEXT-0008.yaml), and [lifetime](../../../spec/contracts/C-CONTEXT-0012.yaml).
+- [Vertical evidence](../../../spec/tests/T-CONTEXT-0003.yaml): Module, Runtime, React, Vue 3, Vue 2 and Web Component scope journeys.
 
-- **tree-based** (logical tree; WC uses DOM tree)
-- **provider → consumer** (single direction in v0)
-- **nearest provider wins**
-- **setup-only** for subscription intent
-- **runtime-only** for reads and updates
-- **JSON-serializable object values** in v0
+Context carries component-to-component values in the nearest self-inclusive provider scope. A subscribed consumer may update that scope; a provider may `update` its own value without subscription. Reads and `tryUpdate` retain their separate subscription requirements. Host ownership determines ancestry; DOM containment is not a universal rule.
 
-These contracts define semantics and phase rules. Integration tests may verify composition (e.g. context → rule → style), but MUST NOT introduce new semantics beyond referenced contracts.
-
----
-
-## Contract Set (v0)
-
-- `with-tree.v0.md`
-- The normative contract for v0 context covers:
-- ContextKey identity and debug name
-- provider resolution (nearest wins)
-- provide/subscribe/trySubscribe phase rules
-- read/tryRead/update/tryUpdate phase rules
-- tree rebind behavior (no notifications in v0)
-- value constraints and portability scope
-- subscription callback semantics
-- error model (codes or equivalent)
-
-- `with-tree.v0.impl-notes.md`
-- Non-normative implementation notes include:
-- recommended internal data structures
-- validation and performance tradeoffs
-- suggested guard strategy and error codes
-
----
-
-## Naming Convention
-
-- Contract docs: `<topic>.v0.md` or `<topic>.v0.impl-notes.md`
-- Contract tests (recommended): `packages/<pkg>/test/contract/context.<topic>.v0.contract.test.ts`
-
----
-
-## Out of Scope (v0)
-
-- Bidirectional/peer communication channels outside context
-- Connection-change notifications (e.g. onConnected/onDisconnected callbacks)
-- Compiler-stage portability and AST-based extraction
-- Host-specific capabilities (e.g. `def.host`)
-
----
-
-## Related Contracts
-
-Context often composes with:
-
-- `rule` (when/then evaluation)
-- adapter realization contracts (host tree representation, lifecycle)
+`with-tree.v0.md` retains the readable topic layout and examples. `with-tree.v0.impl-notes.md` is non-normative implementation background. Connection-change notifications and compiler-stage extraction remain outside this slice.
--- a/internal/contracts/context/with-tree.v0.md
+++ b/internal/contracts/context/with-tree.v0.md
@@ -1,6 +1,6 @@
 # internal/contracts/context/with-tree.v0.md
 
-> Status: Draft – implementation-ready (contract-first) This contract specifies Proto UI **context**: tree-based provider resolution, setup-only subscription intent, runtime reads and updates, and strict value constraints for v0 portability.
+> Legacy readable projection. Current draft authority: [M-CONTEXT-0001](../../../spec/modules/M-CONTEXT-0001.yaml), [C-CONTEXT-0001](../../../spec/contracts/C-CONTEXT-0001.yaml) through [C-CONTEXT-0012](../../../spec/contracts/C-CONTEXT-0012.yaml). These entities supersede this prose; see [the index](./README.md) for host capability and evidence links.
 
 ---
 
@@ -10,7 +10,7 @@
 
 Context provides:
 
-- A **tree-based** communication channel between components (provider → consumer).
+- A **tree-based** communication channel between components within a provider-owned scope (subscribed consumers may update).
 - Setup-only **subscription intent** with runtime **reads** and **updates**.
 - Deterministic provider resolution: **nearest provider wins**.
 - v0 portability constraints: context values are JSON-serializable **objects**.
@@ -29,8 +29,8 @@ Context provides:
 - **ContextKey<T>**: a unique token (symbol-like) identifying a context channel.
 - **Provider**: a component instance that provides a value for a ContextKey.
 - **Consumer**: a component instance that subscribes/reads/updates a ContextKey.
-- **Logical tree**: the runtime component tree (for WC it matches DOM tree).
-- **Nearest provider wins**: the consumer binds to the closest ancestor provider for the key.
+- **Logical tree**: host-translated logical ownership; Web Component derives it from its supported DOM ownership boundary.
+- **Nearest provider wins**: the consumer binds to the closest provider, starting with the consumer itself for the key.
 
 ---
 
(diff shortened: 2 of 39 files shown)


cyjin-yl (changes requested):
Reviewed exact head `7a0e7ac7267e4971b153dba93687660522d2664b`. Repository CI, release lanes, docs preview, Vercel, and DCO are green, and there were no prior `cyjin-yl` reviews/comments or unresolved review threads. I found two blocking source-of-truth / host-contract mismatches:

1. **`C-CONTEXT-0004` still contradicts the self-inclusive resolution this PR catalogs and tests.** `M-CONTEXT-0001-E` says resolution is self-inclusive, the rewritten legacy projection now says “nearest self-inclusive provider”, and `ContextCenter.resolveProvider()` starts at the consumer itself. However, normative criterion `C-CONTEXT-0004-B` still says tree-shaped resolution “must choose the nearest ancestor provider”. At the same time `M-CONTEXT-0001` declares that it satisfies `C-CONTEXT-0004`. That leaves the new catalog internally contradictory about whether a provider is in its own resolution scope. Please reconcile the normative contract (including an explicit revision) with the intended self-inclusive behavior, and keep executable evidence for provider-as-self resolution. The implementation/tests in this head strongly indicate that self-inclusive is the intended result.

2. **The new opaque-identity HC is broader than the implementation’s truthiness assumptions.** `ContextInstanceToken` is currently `unknown`, and `HC-CONTEXT-IDENTITY-0001` describes an opaque token without restricting it to an object/truthy domain. But `ContextCenter.resolveProvider()` traverses with `while (cur)`, and `ContextModuleImpl.dispose()` removes records only under `if (self)`. A valid non-null opaque token such as `0`, `false`, or `''` can therefore be stored by `provide()`, yet fails self-resolution and is not removed on terminal disposal. That violates the newly cataloged identity/lifetime guarantees even though all four official adapters happen to use object tokens. Please either (a) constrain and validate the HC/type to the actual non-null object token domain used by official adapters, or (b) make resolution/disposal use `null` as the only sentinel (`cur !== null`, `self !== null`) and add a regression for a falsy non-null token.

Once these two boundaries are made consistent across contract/HC/implementation/tests, I can re-review the new head.

cyjin-yl (changes requested):
Reviewed exact head `fb989afced494c3c281bd8ee2e0d1cfd214fec6c`. The two blockers from my previous review are materially addressed inside Context itself: provider resolution is now self-inclusive in the normative contract, and Context traversal/disposal now uses null as the absence sentinel. Current repository CI, release lanes, docs preview, and DCO are green. Two integration blockers remain before this widened identity contract is coherent end-to-end:

1. **`ContextModuleImpl.resolveScope()` still loses an explicit `undefined` consumer.** `HC-CONTEXT-IDENTITY-0001-B` now explicitly reserves only `null` for absence, and this follow-up itself tests `undefined` as a valid opaque token. But `resolveScope(key, consumer?)` computes `consumer ?? this.tryGetSelfToken()`, so an explicitly supplied `undefined` identity is treated as if the argument were omitted and resolution silently falls back to the port owner's token. That returns the wrong provider whenever the explicit consumer and port owner have different ancestry. Please distinguish omitted consumer from an explicit `undefined` token (or, if product intent changes, narrow the HC/test domain consistently). Add a regression where the port owner and an `undefined` consumer resolve to different provider scopes so the fallback cannot pass accidentally.

2. **The composed Scroll integration still assumes Context identities are truthy and uses strict equality.** `packages/modules/scroll/src/impl.ts::resolveComposedChrome()` rejects `anatomyScope` with `if (!anatomyScope)`, rejects the resolved scope with `if (!scope)`, and compares identity with `scope !== anatomyScope`. Under the new HC, `0`, `false`, and `''` are valid non-null identities, and `NaN` is also exercised by this PR; all of those can therefore resolve correctly in Context but fail to create the composed scrollbar binding. The equality check also disagrees with the SameValueZero identity semantics used by the provider maps for `NaN`. Please make this integration use null-only absence plus the same identity equality semantics, and add regression coverage through the composed Scroll path for at least a representative falsy token and `NaN` (or the full parameterized domain already used for Context).

Both findings are present on this exact head and are already represented by unresolved review threads. Once those integration boundaries are fixed and the threads are resolved, I can re-review the new head.

cyjin-yl (changes requested):
Reviewed exact head `ce2af1d053d5a8295eda79b33d2969a819f4e020` after the integration follow-up. The two blockers from my previous review are materially fixed: the actual Context port now preserves argument count so an explicit `undefined` consumer is distinct from omission, and the composed Scroll path now uses null-only absence with SameValueZero-compatible identity matching; the new parameterized regressions cover the widened opaque-token domain.

One blocking contract/diagnostic mismatch remains on this head:

- **An omitted-owner `resolveScope(key)` still turns an invalid/missing owner identity into an ordinary missing scope.** `ContextModuleImpl.resolveScope()` now does `arguments.length < 2 ? this.tryGetSelfToken() : consumer`, then returns `null` whenever that value is `null`. But `M-CONTEXT-0001-I` requires operations that need missing identity/ancestry to fail diagnostically, and the ordinary facade path already enforces that through `getSelfToken()` (including a dedicated null-instance-token diagnostic). The current `T-CONTEXT-0003-CASE-AVAILABILITY` also claims this diagnostic boundary, yet it only exercises `provide`/`subscribe`, not the privileged `resolveScope` port. Consequently a miswired host cap (`CONTEXT_INSTANCE_TOKEN_CAP` absent or explicitly `null`) is indistinguishable to a privileged caller from a legitimate “no provider for this valid owner” result.

Please make the omitted-consumer path validate the owner identity (`getSelfToken()` or equivalent) before resolution, while preserving explicit `resolveScope(key, null)` as the deliberate absence sentinel if that remains the intended port contract. Add an actual-port regression showing that a missing owner cap and an owner cap containing `null` both fail diagnostically for `resolveScope(key)`, whereas an explicit `null` consumer retains its intentional null result. Then resolve the existing thread for this finding.

Exact-head repository CI, package/release lanes, docs preview and DCO are green, so this is the remaining blocker I can establish on the current head.

cyjin-yl (approved):
Re-reviewed exact head `f2d1a068a281282e5f653d4dfa02fbc9d76e159d` after the owner-capability diagnostic follow-up. The remaining blocker from my prior review is fixed: omitted-consumer `resolveScope(key)` now routes through `getSelfToken()`, so a missing or explicit-null owner identity fails diagnostically, while an explicit `resolveScope(key, null)` remains the intentional absence sentinel. The new actual-port regression also covers missing identity, null identity, missing ancestry, explicit null, and explicit valid consumers.

I rechecked the incremental commit rather than relying on the PR description: it changes only the Context port implementation/comment and focused catalog-boundary tests, and it preserves the previously reviewed explicit-`undefined` consumer behavior and null-only identity semantics. All prior review threads are resolved. Exact-head repository `test`, `type-check`, `lockstep-check`, `release-scan`, CI/Poppy preview, Vercel, and DCO are green, with DCO reporting all commits signed off. I do not find a remaining blocking correctness, contract, lifecycle, dependency, or scope issue on this head.

PR #634, opened 2026-09-09
feat: finish Overlay, Text Control, Image View and Scroll catalog

guangliang2019:
本批次收尾 Overlay、Text Control、Image View、Scroll 的 M／HC／A／T 编目，并修复沿证据链发现的局部漂移。所有实体保持 draft；完成这四项后结束批次，后续架构议题独立评估。

- **Overlay**：落实已接受的同 scope、同 Escape sample 唯一候选策略，覆盖受控同步 reopen 与嵌套 Dialog／Dropdown Root request；Event 仅提供内部 opaque input identity。新增 Portal／modal／layer 三个 HC，修复并发 modal overflow 锁、provider 替换释放和 layer disposer 的 priority／幂等性。
- **Text Control**：补全四 Adapter 与单行／多行编辑证据；旧 lease 回调不能再更新当前值，listener cancellation 遵循 setup-only。
- **Image View**：补全 generation 生命周期及 M／HC／A 映射；为 Vue 2 接入已有 Web physical image／host bridge，验证替换 source、旧 decode、空 source 和卸载；更新依赖、lockfile 与生成的 BOM。listener cancellation 对齐现有 Core 契约。
- **Scroll**：补全 State／Context／Anatomy 依赖及四 Adapter 普通和 composed Thumb 通路；修复替换／detach 后迟到 facts 仍能覆盖当前 session 的问题。

每个切片有独立 commit。边界与后续清单见 `internal/records/2026-09-09-overlay-and-remaining-catalog.zh-CN.md`、`internal/records/2026-09-09-ordinary-catalog-completion.zh-CN.md`。本 PR 不推进 A11y asHook 重设计、Rule Meta、Presence/deprecated State 清理、完整 modal/focus 策略或非 Web 认证。

验证：

- Spec fixture/graph：22 文件、61 项通过；catalog、完整 workspace/docs types、Agent projection、package manifests/BOM 检查通过。
- `pnpm test` 的治理、发布、public-docs、type contracts 通过；非浏览器回归 447 文件、2048 项通过，既有 3 skipped files／34 todo。
- Vue 2 package 及构建依赖通过（35 packages）。
[...]

Issue #621, opened 2026-09-04
[implementation][table] Deliver Base Table Structure and A11y graph slices


cyjin-yl:
Parent proposal: #387
Decision/spec PR: #620

## Goal

Implement the draft passive Base Table structural protocol defined by `D-BASE-TABLE-STRUCTURE-0001`, `C-TABLE-STRUCTURE-0001`, and `T-TABLE-STRUCTURE-0001` after #620 merges.

The implementation must close the current opaque accessibility-relation gap, deliver the Table Structure semantic owner and five Base identities with executable evidence, and add bounded Web Adapter projection without inventing a Table-specific host service or Data Table behavior.

## Fixed semantic boundary

The accepted first slice is:

- planned identities: `P-BASE-TABLE`, `P-BASE-TABLE-CAPTION`, `P-BASE-TABLE-ROW`, `P-BASE-TABLE-HEADER-CELL`, and `P-BASE-TABLE-CELL`;
- one planned `M-TABLE-STRUCTURE-0001` semantic owner;
- Anatomy-ordered zero-based two-dimensional placement with a monotonic per-Row cursor and positive safe-integer `rowSpan` / `columnSpan` with range-based occupancy;
- same-domain non-empty unique `headerKey` matching, placeable targets, row-major-upstream acyclic HeaderCell relationships, multiple ordered row/column header relationships, and strict nearest-domain isolation;
- read-only validity/diagnostic snapshots and recomputation on insert/remove/reorder;
- the generic `M-A11Y-0001` / `HC-A11Y-0001` path;
- no Head, Body, RowGroup, `HC-TABLE`, Collection dependency, or host-specific materialization guarantee.

Sorting, row/cell selection, pagination, virtualization, column resize/reorder, editing, data fetching, keyboard-grid/focus policy, form workflow, layout tables, and shadcn/Brutalist projections remain outside this Issue.

## Delivery graph

Keep each checkpoint in a separate coherent PR.

### Checkpoint A — opaque A11y relationship prerequisite

- Extend the generic A11y IR with opaque same-domain semantic-object references and ordered multi-target relationships consistent with `D-A11Y-PART-RELATIONSHIP-PROJECTION-0001`.
- Preserve current string/State-backed relation behavior and additive Web ownership where still governed.
- Resolve protocol keys inside the owning semantic domain; never expose those keys as host IDs.
- Add runtime and projector evidence for identity stability, multiple targets, missing targets, rematerialization, removal, and cleanup.
[...]

Issue #549, opened 2026-08-26
[prerequisite] Reusable same-domain anatomy-part relationship (#388 prerequisite)


cyjin-yl:
## Goal

Deliver the accepted reusable same-domain anatomy-part relationship prerequisite for the Disclosure/Collapsible/Accordion family under #388.

## Context

Per the #388 maintainer ruling, two prerequisites must be completed before Collapsible/Accordion implementation:

1. **Separate issue**: portable heading-level prerequisite
2. **This issue**: reusable same-domain anatomy-part relationship prerequisite

## Scope

- Define how an anatomy part (Trigger → Content) declares a same-domain opaque relationship
- The relationship must survive L1 detach (logical identity preserved, Web ARIA IDREF removed while detached, restored on rematerialization)
- Tabs already uses a similar Trigger→Content relationship; provide migration evidence
- The relationship must be reusable across Disclosure, Collapsible, and Accordion

## Agent-ready progression

Checkpoint A already fixes this prerequisite's semantics. Agents may implement, validate, respond to review, and integrate it without repeated activation. Accepted exact-head completion automatically unlocks Base Collapsible. This remains prerequisite-only: no Collapsible, Accordion, shadcn, or Brutalist implementation belongs in this slice.

## Acceptance criteria

- [ ] Duplicate/missing keys and zero/multiple matches fail closed without guessing.
- [ ] L1 detach removes the Web IDREF while preserving logical relationship and opaque reserved identity; rematerialization restores the same target identity before reveal.
- [ ] Tabs is migrated as the second real consumer.
- [ ] No host node, host ID, framework key, or raw geometry enters portable Props/State/Context/Expose.
- [ ] Trusted exact-head CI and DCO/provenance pass; an independent reviewer accepts the head and every applicable thread is resolved.

- [ ] Same-domain anatomy-part relationship is defined (spec entity or contract amendment)
- [ ] Tabs migration evidence shows existing relationships working with the new pattern
- [ ] Detached relationship projection is defined (logical identity survives, Web ARIA removed while detached, restored on rematerialization)
- [ ] All existing tests pass

## Agent-forward execution

Current decision: `none for the currently eligible bounded transition`.

This bounded work is open for Agent execution. Agents may continue through research or proposal work, implementation of any direction fixed by applicable authority, validation, review response, and the next eligible dependency under current-user or active standing authorization. Routine claiming, implementation details, CI repair, review, ready-for-review, and merge mechanics are not attended decisions.

[...]

PR #636, opened 2026-09-10
feat(a11y)!: replace def.a11y with asAccessible

guangliang2019:
## Summary

Replace the 0.3 prototype author entry `def.a11y` with `asAccessible()` from `@proto.ui/hooks`, returning the core `AccessibleHandle`. Remove the former property and `A11yDefAPI` type without compatibility aliases. Migrate official consumers, executable evidence, author docs and draft release notes; declare Shadcn's new direct hooks dependency.

Repeated setup calls share one instance handle. Declaration methods remain setup-only even on retained handles, and dynamic facts retain State ownership. This changes the author entry, not accessibility interaction behavior or host projection policy.

## Related context

- Related integration work: #549 (same-domain Anatomy relationships) and #621 (Table / A11y graph). This PR changes their future author entry only; it does not satisfy their relationship or Table acceptance criteria and does not close either Issue. No dedicated existing Issue for the asAccessible migration was found.

- Maintainer-approved follow-up to #634; independent architecture slice.
- Draft authority: `C-A11Y-0001`, `M-A11Y-0001`, `HC-A11Y-0001`, `T-A11Y-0001`; lifecycle statuses remain unchanged, no admission proposed.
- New criteria: `C-A11Y-0001-AUTHOR-HANDLE`, `C-A11Y-0001-DECLARATION-LIFETIME`, `C-A11Y-0001-DOMAIN-BOUNDARY`.
- Evidence: `T-A11Y-0001-CASE-ACCESSIBLE-HANDLE` plus migrated existing Runtime and four Adapter tests.
- #625 remains a separate open direct-reference transport change. Its eventual integration must reconcile the same Module; this PR does not expose its internal mutation port.

## Scope boundary

The maintainer approved the name, shared setup handle and direct breaking migration on the 0.3 main line. No 0.2 backport, compatibility period, multiple-object author API, arbitrary host target, runtime mutation, disposal API or new action execution path. State, Focus, Anatomy and interaction modules retain their ownership.

## Source-of-truth alignment

- [x] Applicable spec entities were traced before implementation.
- [x] Criteria, test mappings, executable tests and reader projections changed together.
- [x] No placeholder identity or lifecycle promotion was introduced.

## Contribution provenance

No external implementation was copied. A coding agent traced existing repository sources, implemented the migration, updated specs/tests/docs and ran local verification. The human maintainer reviewed and approved the design and naming; independent implementation review remains pending. No external private or third-party source was supplied for this task.

## DCO

The commit includes the configured original contributor's matching Signed-off-by trailer. GitHub DCO verification remains a live check.

## Prototype website preview

- [x] A new Prototype page is not applicable: existing identities and visible behavior are preserved; bilingual author documentation is updated.
- Existing Shadcn controls browser regression consumes the migrated implementations and passed in the Web Component, React and Vue scope covered by that fixture.
- No new demo orchestration or consumer-owned behavior was introduced.

## Validation

- New handle + A11y Runtime + Vue 2 heading tests: 3 files, 39 tests passed.
- Full nonbrowser Vitest run: 455 files / 2164 tests passed, with one catalog failure caused by a new source not yet being Git-tracked. After indexing it, catalog evidence integrity passed 3/3. Existing 3 skipped files and 34 todo remain.
- Public build: 42 packages passed initially; Shadcn exposed a missing direct hooks dependency. After fixing manifest/lock/BOM, Shadcn's complete 11-package closure built successfully.
- Full workspace and docs types passed (195 Astro files, zero diagnostics); type contracts passed.
- Spec authoring against `e15a5adf`, prototype catalog, generated workspace/Agent projections, Agent checks, public-docs tests, all public manifests, release assets and all nine bundle budgets passed. Web Component measured 74767/76000 gzip bytes locally.
- Shadcn controls real-browser regression: 5/5 passed with a 60-second hook timeout.
- Not run: full browser matrix and isolated installed-consumer/release smoke. Required CI must validate the published PR head; local evidence is not complete assistive-technology certification.

### CI follow-up

The first CI test job failed because Shadcn's preserved README omitted the newly declared `@proto.ui/hooks` production dependency. Added it to Related Internal Packages; `node --test scripts/release/test/readme-metadata.test.mjs` now passes 4/4 and all 43 manifest checks pass. The initial public build, type check, release stage and installed consumer jobs passed; fresh CI must verify the follow-up head.

### Review follow-up: registration collision

Implemented the review's explicit-rejection option in Runtime registration, covering every mode before once/configure reuse. A name already registered by an authored hook cannot be reused by a privileged hook, or vice versa; a diagnostic identifies the conflicting name. Same-kind reuse remains unchanged. This deliberately rejects the mixed-name composition rather than introducing a new public lookup/namespace API; rename the authored hook to compose both.

Both requested call orders failed before the fix and pass now. Runtime/Core/Hooks regression: 64 files, 311 tests passed (existing 3 skipped files / 34 todo). Workspace types, spec authoring, catalog integrity and generated projection checks passed. New exact-head CI and independent review remain pending.

### DIFF
--- a/apps/www/src/content/docs/en/build/prototypes/writing-a-custom-primitive-prototype.md
+++ b/apps/www/src/content/docs/en/build/prototypes/writing-a-custom-primitive-prototype.md
@@ -35,7 +35,7 @@ The current implementation includes:
 - `asFocusable()` for `focused`, `focusVisible`, and the focus method;
 - `def.event.on()` for pointer routes and `press.commit`;
 - `def.expose.state()`, `def.expose.method()`, and `def.expose.event()` for outward surfaces; and
-- `def.a11y.*` for Button role, name, state, and action.
+- `asAccessible()` for Button role, name, state, and action.
 
 The former `def.state.fromInteraction()` example no longer describes the current Button implementation and must not be used as the example for this guide.
 
--- a/apps/www/src/content/docs/zh-cn/build/prototypes/writing-a-custom-primitive-prototype.md
+++ b/apps/www/src/content/docs/zh-cn/build/prototypes/writing-a-custom-primitive-prototype.md
@@ -35,7 +35,7 @@ Button 的两个官方 authoring entries 是：
 - `asFocusable()` 提供 `focused`、`focusVisible` 与 focus method；
 - `def.event.on()` 承接 pointer 与 `press.commit`；
 - `def.expose.state()`、`def.expose.method()` 和 `def.expose.event()` 提供 outward surface；
-- `def.a11y.*` 声明 Button 的 role、name、state 与 action。
+- `asAccessible()` 声明 Button 的 role、name、state 与 action。
 
 旧的 `def.state.fromInteraction()` 例子已经不代表当前 Button 实现，不应继续作为这篇指南的示例。
 
--- a/internal/records/2026-09-10-as-accessible-migration.zh-CN.md
+++ b/internal/records/2026-09-10-as-accessible-migration.zh-CN.md
@@ -1,0 +1,20 @@
+# asAccessible 作者入口迁移
+
+用户已接受在 0.3 main 直接用 `asAccessible()` / `AccessibleHandle` 替代 `def.a11y` / `A11yDefAPI`，无兼容期、无旧入口别名、不回补 0.2。
+
+统一入口返回当前逻辑实例的唯一声明 handle；setup 重复获取不重置声明，单值字段或同 key 声明最后一次生效，tree 按字段合并。声明和保存 handle 上的方法都限 setup，动态值由已有 State 驱动。没有新增 dispose、通用 runtime mutation、多对象创建或 host target API。
+
+A11y 拥有可投影的语义表达，State、Focus、Anatomy、Trigger/Event 和领域模块保留事实、行为与目标解析所有权。调用 asAccessible 不构成完整无障碍认证，也不安装键盘或焦点策略。
+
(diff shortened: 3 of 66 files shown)


cyjin-yl (changes requested):
Reviewed exact head `bade76ab9180b5a1e4f244d986a7928221154fe0`. One blocking compatibility/runtime-identity issue remains.

`packages/hooks/src/as-accessible.ts` registers the privileged hook under the public string name `asAccessible`. `definePrivilegedAsHook()` passes that same string to `rt.register()`, while public `defineAsHook()` also accepts arbitrary non-empty names and registers them in the same runtime. `attachAsHookRuntime()` stores once-hook state in `Map<string, ...>` keyed only by `name`; on a collision it returns `skip` without checking privileged/unprivileged identity. Therefore a consumer `defineAsHook({ name: 'asAccessible', ... })` can occupy the slot first and cause the built-in call to reuse the consumer result as an `AccessibleHandle`, or the built-in can occupy it first and silently suppress the consumer hook. The existing unresolved automated thread identifies the same behavior, and the exact-head runtime still has it.

Please make the privileged registration identity non-colliding (for example, a private/namespaced key distinct from the public trace/display name), or make registration reject a privileged/unprivileged identity collision before reuse. Add executable regressions for both call orders: consumer hook first then `asAccessible()`, and `asAccessible()` first then consumer hook. The acceptance criterion is that the built-in can never return a consumer result as `AccessibleHandle`, and adding this built-in must not silently suppress an otherwise valid consumer hook.

cyjin-yl (approved):
Reviewed exact head `b223a25747495cbb9ca52c83558912ce04d4e716` after the requested collision fix. The new runtime guard rejects authored/privileged reuse of the same registration name before once/configure reuse, so neither call order can return a consumer result as `AccessibleHandle` or silently suppress the other hook. The added regression covers both authored-first and privileged-first orders, and the contract/docs were updated to match the implemented failure mode. Current repository CI and DCO are green on this head, and the previously blocking review thread has been reconciled as fixed. No remaining blocking issue found on the reviewed head.

PR #661, opened 2026-09-18
fix(module-overlay): compensate removed scrollbar in web scroll lock

cyjin-yl:
## Problem

Dogfooding the docs site (`apps/www`): opening any modal overlay — dialog, search palette — on a scrollable page removes the viewport scrollbar (`overflow: hidden` on `body`) with no compensation, so the whole page visibly shifts to the right every time an overlay opens and back when it closes.

## Root cause

All four adapters delegate modal scroll-locking to `createWebOverlayModal` in `@proto.ui/module-overlay` (`packages/modules/overlay/src/web/modal-lock.ts`). The lock only snapshotted/restored `overflow`; it never compensated the removed scrollbar.

## Fix

In the shared web scroll-lock realization:

- On first lock, measure the viewport scrollbar width (`innerWidth - documentElement.clientWidth`, so overlay-scrollbar platforms and non-scrollable pages measure `0` and are untouched).
- Add that width on top of the body's current computed `padding-right`, snapshotting the original inline value and priority exactly like the existing `overflow` snapshot.
- On last unlock, restore both properties. Reference counting and cross-owner protection are unchanged.

## Verification

- New `packages/modules/overlay/test/modal-lock.test.ts`: 5 cases (basic compensation, stacking on existing padding, no-scrollbar no-op, multi-owner reference counting, inline overflow value+priority restore) — 6/6 overlay tests pass.
- `vitest run packages/adapters -t overlay`: 19 passed.
- `tsc -p tsconfig.workspace.json --noEmit`: clean.
- Web-component adapter gzip: 75,664 → 75,855 bytes (+191), measured with the canonical local budget script.

Signed-off-by: 255doesnotexist <255doesnotexist@users.noreply.github.com>


cyjin-yl:
Additional evidence gathered after opening:

**Live-site reproduction (before this fix)**: on `www.proto-ui.com/en/ui-libraries/base/dialog/` (scrollable page, classic 15px scrollbar), opening the dialog sets `body { overflow: hidden }` with `padding-right` untouched (`""`), so page content shifts right by the scrollbar width. Confirmed with headless Chrome measurement; zero console errors on live.

**Local e2e caveat**: `astro build && astro preview` on Windows currently fails to hydrate docs demo islands (45 `[AsHook] no active setup context` page errors; dialog never reaches `aria-modal=true`, so the modal lock never engages). This reproduces identically on unmodified main (`ddac15da`) and this branch, so it is a pre-existing local-preview issue, unrelated to this change — filing it separately. It means the docs-site e2e verification of this fix has to happen on a real deploy; the unit tests in `modal-lock.test.ts` cover the compensation logic itself.

cyjin-yl:
更正上一条：本账号（cyjin-yl）在仓库只有 push+triage 权限，无 admin，无法绕过分支保护。需要广亮或 baiyucore 在 GitHub 上提交 Approve 后才能合入。

guangliang2019 (approved):
Reviewed exact head `ec62b1a9ea7b8a62d49b18a51e209923bf474700`.

